Issue #14, opened 2017-12-04
Add Electronic bank statements and transactions


willemstuursma:
* https://c3.twinfield.com/webservices/documentation/#/ApiReference/Transactions/BankStatements
* https://c3.twinfield.com/webservices/documentation/#/ApiReference/Transactions/BankTransactions

modernmindsnl:
has something already been made to GET Electronic Bank Statements and not only Add them?

willemstuursma:
@modernmindsnl Sorry, this is not implemented and we've no plans to implement this yet.

We're perfectly willing to take PR's for this.

You could start by adding a `get` method to `ElectronicBankStatementApiConnector` and adding a mapper to transform the XML from Twinfield into objects.

PR #18, opened 2017-12-04
#14 Add Electronic Bankstatements and -transactions

willemstuursma:


willemstuursma:
@Daanvm do you need the mapper too?

src/ElectronicBankStatement.php
```diff
@@ -0,0 +1,215 @@
...
+    /**
+     * @param array|ElectronicBankStatementTransaction[] $transactions
+     */
+    public function setTransactions(array $transactions): void
+    {
+        Assert::allIsInstanceOf($transactions, ElectronicBankStatementTransaction::class);
+        $this->transactions = $transactions;
+
+        foreach ($transactions as $transaction) {
+            $this->closevalue = $this->closevalue->add($transaction->getValue());
```
Daanvm: This only works when `$this->closevalue` is already set to the start value in `setStartvalue()`. Maybe add an assertion that `$this->startvalue` is not empty?

And setting the transactions a second time will overwrite the transactions, but the close value won't be reset. I think a `$this->closevalue = $this->startvalue` before the foreach-loop will fix this.

Daanvm:
Note that the transactions from this PR are not the [Bank Transactions from the documentation](https://c3.twinfield.com/webservices/documentation/#/ApiReference/Transactions/BankTransactions) you linked to in issue #14.

We need both, so maybe we should add a new issue for the Bank Transactions.

PR #20, opened 2017-12-05
Issue/19 refactor client

willemstuursma:
This simplifies some of the class usage.

* Removes the `Service` class. Makes lots of methods on the `Login` class private.
* Stuff has been moved around from `Service` to `SoapClient` and `Login` classes.
* ApiConnectors are not simplified. They take a `Login` object, and just have to send a `DomDocument`. All error handling is moved up and generalized. 
* Retry mechanism has been removed. 
* All errors now result in `Exception`s and no more true / false checking needed. 

This breaks the Finder though, because it uses a custom WSDL. Have to look into that.


willemstuursma:
I guess every API connector should indicate to which SOAP endpoint it wants to send its document. 

https://c3.twinfield.com/webservices/documentation/#/GettingStarted/WebServicesOverview

I'll make some changes for that.

willemstuursma:
I'll merge this now, and then I'll make a separate issue for the different WSDL files.

PR #23, opened 2017-12-06
Issue/21 banktransaction objects

willemstuursma:
[wip]


willemstuursma:
OK @Daanvm I added the base class and three classes for the lines.

Lots of shared fields were hoisted out to traits. 

Matching is not yet implemented

willemstuursma:
Note that this also implements `Money` value objects for most fields documented as `money`.

src/BaseTransactionLine.php
```diff
@@ -2,6 +2,10 @@
 
 namespace PhpTwinfield;
 
+use PhpTwinfield\Enums\LineType;
+use PhpTwinfield\Transactions\TransactionLineFields\ValueFields;
+use SebastianBergmann\Diff\Line;
```
Daanvm: I think this import is a mistake.

src/ElectronicBankStatement.php
```diff
@@ -161,7 +142,11 @@ public function setTransactions(array $transactions): void
         $this->closevalue = $this->startvalue;
 
         foreach ($transactions as $transaction) {
-            $this->closevalue = $this->closevalue->add($transaction->getValue());
+            if ($transaction->getDebitCredit() == DebitCredit::CREDIT()) {
+                $this->closevalue = $this->closevalue->add($transaction->getValue());
+            } else {
+                $this->closevalue = $this->closevalue->subtract($transaction->getValue());
```
Daanvm: Shouldn't you always `add()` the value, because the value is negative in case of a debit transaction?

src/Mappers/TransactionMapper.php
```diff
@@ -93,12 +102,12 @@ public static function map(string $transactionClassName, Response $response): ar
...
 
                 $transactionLine
-                    ->setType($lineElement->getAttribute('type'))
+                    ->setType(new LineType($lineElement->getAttribute('type')))
                     ->setId($lineElement->getAttribute('id'))
                     ->setDim1(self::getField($transaction, $lineElement, 'dim1'))
                     ->setDim2(self::getField($transaction, $lineElement, 'dim2'))
-                    ->setDebitCredit(self::getField($transaction, $lineElement, 'debitcredit'))
-                    ->setValue(self::getField($transaction, $lineElement, 'value'))
+                    ->setValue(Money::EUR(100 * self::getField($transaction, $lineElement, 'value')))
```
Daanvm: Transactions can be in other currencies also. Not sure if we do have to support it yet, but in that case we should at least assert that the transactions currency is `EUR`.

src/Mappers/TransactionMapper.php
```diff
@@ -2,10 +2,15 @@
```
Daanvm: Can we use this 'general' `TransactionMapper` also to get the new `BankTransactions` or do we need to create a new mapper if we would like to get them from Twinfield?

src/BankTransaction.php
```diff
@@ -0,0 +1,141 @@
...
+     */
+    private $transactions = [];
+
+    /**
+     * The bank transaction origin.
+     * Read-only attribute.
+     *
+     * @return mixed
+     */
+    public function getOrigin()
```
Daanvm: Without a setter function for read-only attributes there's also no way for a mapper to set the attributes. But there's currently no mapper to get these kind of transactions from Twinfield, right? So the setters aren't really needed yet.

willemstuursma: Yes, my thoughts exactly.

src/BaseTransactionLine.php
willemstuursma: Thanks.

src/ElectronicBankStatement.php
willemstuursma: The value from a transaction is always a positive amount, the debit credit field is used to signal the sign.

willemstuursma:
```
    public function getValue(): Money
    {
        /*
         * Always return the absolute value, Twinfield uses "debitcredit" fields instead of signs.
         */
        return $this->value->absolute();
    }
```

src/Mappers/TransactionMapper.php
willemstuursma: Yes, I think we should handle multi-currency when we get there.

src/Mappers/TransactionMapper.php
willemstuursma: I can give it a try, if you provide me with an XML file I will create / adapt the mapper.

Daanvm: I used the Request and Response examples from the documentation, in this case at the bottom of [the Bank transactions documentation](https://c3.twinfield.com/webservices/documentation/#/ApiReference/Transactions/BankTransactions), under the 'XML Example' header.

PR #24, opened 2017-12-06
Support many WSDLs. 

willemstuursma:
Refactors basically all ApiConnectors, introduces services.


Daanvm:
Nice!
The one thing this branch does not support yet is using the `Finder` service to get a list of items and use the `ProcessXML` service to create a new item in the same ApiConnector. But we can always add that when we need it.

PR #28, opened 2017-12-15
Private to public fields

kelvinsouer:
Made private data fields public. So that it is easier to work with in our current system


willemstuursma:
It would be better to create getters / setters if you need them, but ala.s

PR #34, opened 2017-12-23
don’t create empty elements

alexjeen:
if you don’t specify a < code > element, twinfield will automatically generate a new code according to the administration it’s mask. If you specify < code > Twinfield will actually say that code is required. If you don't pass in the element at all the code will be automatically generated.

getFirstFreeCode is unreliable as administrations can use different masks. It will also not work on administration with multiple (legacy) ranges

I have a lot of fixes and new functionalities for this library. I also improved the OAuth library because right now it is dependant on a $_SESSION. If you use something else for your intermediate token storage (for instance a framework for session management) this won't work

Also you maybe should remove the password authentication part from the documentation because this is officially deprecated by Twinfield

I also implemented OAuth V2.0 for this library (this just in by Twinfield). Next week I will make some new pull requests if I get around to it

### DIFF
--- a/src/DomDocuments/CustomersDocument.php
+++ b/src/DomDocuments/CustomersDocument.php
@@ -67,14 +67,16 @@ class CustomersDocument extends \DOMDocument
         foreach ($customerTags as $tag => $method) {
 
             // Make text node for method value
-            $node = $this->createTextNode($customer->$method());
+            if($customer->$method()) {
+                $node = $this->createTextNode($customer->$method());
 
-            // Make the actual element and assign the node
-            $element = $this->createElement($tag);
-            $element->appendChild($node);
+                // Make the actual element and assign the node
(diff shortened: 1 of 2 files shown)


willemstuursma:
Thanks. Could you update the test too?

Looking forward to your other pull requests too.

willemstuursma:
Thanks @alexjeen. However, the build is still failing. 

You could fix it, then I will merge your changes.

You can see the problem [here](https://travis-ci.org/php-twinfield/twinfield/jobs/321063293).

alexjeen:
I know, but I have a local PHP version (MAMPP) of PHP7.0, but I need PHP 7.1 to do the testing correctly (just changed the XML file blindly)

I will install it after christmas and then fix the test.

willemstuursma:
Thanks, and merry Christmas! 🎄

PR #41, opened 2017-12-29
Implement OAuth version 2 (base)

alexjeen:
As said, I have implemented OAuth 2 now. Works pretty well with the existing library. 

A list of todo's before it can be merged:

1. Write tests
2. I would propose we store office in the Connection class. See my second comment why
3. Test if the refreshing of tokens work (I didn't test that)

Nice to have would also be to build in the browse codes that Twinfield uses. This allows you to get open invoice list etc.

As I am very busy the coming year, I don't have time to write any tests and to do further integration, sorry for that, I hope this helps. I actually did this in my holiday :)

I have also involved the guys of Picqer in this library, I know they have some further improvements for this library.


alexjeen:
Regarding 2. 

The office is really central in all communicating with Twinfield so I think we should store it centrally. So you should not pass it into all the ApiConnectors. We should store it in Connection class and make it easy to get / set. It should also be required for the library to work because OAuth v2.0 does not work without this code.

Furthermore you wrap it into an Office class now. This will cause some problems. Because if you have a level 4 user in Twinfield authentication this user can for instance not get a list of Offices (because this is restricted in the permissions). This way you have no information to instantiate your Office class.

If you want to communicate with multiple offices, you can easily do a call to $connection->setOffice('x') and the next calls will use the new office.

This is just a proposal but in my day to day work with Twinfield I think it would be cumbersome to keep on passing in the same value.

alexjeen:
Also, I have added myself as the composer author, you should do the same I think. 
And after this is integrated the documentation should be updated with the example.

willemstuursma:
I want to make some changes to have different objects for configuration and different strategies for logging in, instead of the generic `Config` and `Connection` class.

willemstuursma:
I've created the class to authenticate using an access token. Once I have my client ID and secret from Twinfield I will test if it works.

Now, how you get the access token. 

I don't think _getting_ an access token belongs in this library.

The best way would be to create a Twinfield provider for the OAuth2 client, like the [Mollie provider](https://github.com/mollie/oauth2-mollie-php). 

Then we won't have to put any session management, URL management, expiry management, redirection and validation in this library and you can properly integrate in your own application.

alexjeen:
I think that's a very good idea, I only needed to do some minor tweaks to get it to be compatible with the GenericProvider.

willemstuursma:
We could also include the provider as part of this library, but it needs to go into a separate class.

willemstuursma:
We'll move this further in #75.

Issue #52, opened 2018-01-09
Sending a sales invoice returns a null response


Waterdhavian:
I create a sales invoice object and send it to twinfield classic,
If i go to my twinfield classic > invoices section then i can see the invoice that i just created from laravel.
I use the code below to send my invoice.

```
$result = $invoiceApiConnector->send($invoice);
dd($result);
```

I would expect to get a complete invoice object back, according the twinfield API docs the result is an invoice object which includes the generated invoice_number which i need to generate a PDF in laravel.

Am i wrong in assuming i would get a response back?
  

This my full code, using the facade by willemo:
```
$officesApiConnector = \Twinfield::get('Office');
		
$offices = $officesApiConnector->listAll();
$office = $offices[0];
		
$customerApiConnector = \Twinfield::get('Customer');
$customer = $customerApiConnector->get('1001', $office);
		
$invoiceApiConnector = \Twinfield::get('Invoice');
		
$line = new \PhpTwinfield\InvoiceLine();
$line->setQuantity(1.00);
$line->setArticle(0);
$line->setUnitsPriceExcl(100.00);
$line->setUnits(1);
$line->setVatCode("VN");
$line->setDim1(8020);
$line->setDescription("Test");
$line->setAllowDiscountOrPremium(false);
		
$invoice = new \PhpTwinfield\Invoice();
$invoice->setCustomer($customer);
$invoice->setOffice($office);
$invoice->addLine($line);
$invoice->setStatus("concept");
$invoice->setPaymentMethod("bank");
$invoice->setInvoiceType("FACTUUR");
$invoice->setInvoiceDate("20180104");
$invoice->setCurrency("EUR");
$invoice->setBank("BNK");
$invoice->setPeriod('2018/01');
$invoice->setDueDate(\Carbon\Carbon::now()->addMonth());

// Send it to Twinfield.
$result = $invoiceApiConnector->send($invoice);
		
dd($result);
	
return view('index');
```

willemstuursma:
`public function send(Invoice $invoice): void` is a void function and does not return anything. 

You need to use `\PhpTwinfield\ApiConnectors\InvoiceApiConnector::get()` to retrieve the invoice again.

Waterdhavian:
```
public function get(string $code, string $invoiceNumber, Office $office)
    {
        // Make a request to read a single invoice. Set the required values
        $request_invoice = new Request\Read\Invoice();
        $request_invoice
            ->setCode($code)
[...]

PR #53, opened 2018-01-09
Return the mapped invoice object from twinfield

Waterdhavian:
I had to change the Invoicemapper from isset to !empty.
Otherwise it will execute the setPerformanceType, which gets an empty string as input but it expects an ENUM of type PerformanceType

Fix for issue #52

### DIFF
--- a/src/ApiConnectors/InvoiceApiConnector.php
+++ b/src/ApiConnectors/InvoiceApiConnector.php
@@ -52,16 +52,16 @@ class InvoiceApiConnector extends ProcessXmlApiConnector
      * @param Invoice $invoice
      * @throws Exception
      */
-    public function send(Invoice $invoice): void
+    public function send(Invoice $invoice)
     {
-        $this->sendAll([$invoice]);
+        return $this->sendAll([$invoice]);
     }
 
     /**
      * @param Invoice[] $invoices
      * @throws Exception
      */
-    public function sendAll(array $invoices): void
+    public function sendAll(array $invoices)
     {
         Assert::allIsInstanceOf($invoices, Invoice::class);
 
@@ -73,7 +73,9 @@ class InvoiceApiConnector extends ProcessXmlApiConnector
                 $invoicesDocument->addInvoice($invoice);
             }
 
-            $this->sendDocument($invoicesDocument);
+            $response = $this->sendDocument($invoicesDocument);
+			
+            return InvoiceMapper::map($response);
         }
     }
 }
--- a/src/ApiConnectors/InvoiceApiConnector.php
+++ b/src/ApiConnectors/InvoiceApiConnector.php
@@ -55,7 +55,8 @@ class InvoiceApiConnector extends ProcessXmlApiConnector
      */
     public function send(Invoice $invoice): Invoice
     {
-        return reset($this->sendAll([$invoice]));
+        $result = $this->sendAll([$invoice]);
+        return reset($result);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Mappers/InvoiceMapper.php
```diff
@@ -104,7 +104,7 @@ public static function map(Response $response)
             foreach ($lineTags as $tag => $method) {
                 $_tag = $lineDOM->getElementsByTagName($tag)->item(0);
 
-                if (isset($_tag) && isset($_tag->textContent)) {
+                if (isset($_tag) && !empty($_tag->textContent)) {
```
Waterdhavian: I had to change this to prevent errors when setting the performance type because it expects an ENUM but receives an empty string.

Waterdhavian:
No idea why the builds are failing, i don't have much experience with CI and i could not find any info inside Travis why the builds are failing.

alexjeen:
Its because a test is failing:

https://api.travis-ci.org/v3/job/326887227/log.txt

See all the way at the bottom

alexjeen:
3fa62481-3bdb-47a5-accc-fa803bf4471e

Waterdhavian:
Haha thank you @alexjeen!

willemstuursma:
I've made some generalizations in the Invoice mapper.

The only thing I that I have not tested yet is if the `\PhpTwinfield\ApiConnectors\InvoiceApiConnector::sendAll` method which should return an array of `Invoice` objects. 

@Waterdhavian can you make the changes to make that work? It sends the Invoices wrapped another XML tag so you probably need to iterate over  a tag in `\PhpTwinfield\Response\Response::getResponseDocument()` and then pass that to the `InvoiceMapper`, which should then take a `DOMDocument` as a constructor argument instead of the `Response object`.

Waterdhavian:
@willemstuursma i will see if i can do that tomorrow. I am traveling today.

Waterdhavian:
I have fixed the sendAll feature, had to make a change in the baseMapper because perfomancedate is an optional date which is empty a lot of the times. 

I cannot change the "Changes Required", or i don't know how to do that ;)

src/ApiConnectors/InvoiceApiConnector.php
```diff
@@ -73,7 +77,11 @@ public function sendAll(array $invoices): void
                 $invoicesDocument->addInvoice($invoice);
             }
 
-            $this->sendDocument($invoicesDocument);
+            $response = $this->sendDocument($invoicesDocument);
```
willemstuursma: Did you test if this works correctly? 

The invoices are sent in chunks (of 250 objects). However, you call the `InvoiceMapper::map()` only once per sent document. 

Can you test sending multiple invoices and getting multiple mapped invoices back?

Waterdhavian: Hello Willem,

I will run some tests today / tomorrow and get back you :)

Waterdhavian:
@willemstuursma So you want me to write a generator to iterate over all the invoice objects that are sent to the sentall function and yield back an array with mapped invoice objects?

Also what do you mean by: 
"The invoices are sent in chunks (of 250 objects). However, you call the InvoiceMapper::map() only once per sent document."

As far as i understood every invoice has it owns invoice object, so sending a chunk of 250 invoices would result in an array consisting of 250 mapped invoice objects. 

I am just asking because i would like to make a good contribution to this project which is in line with the rest of the coding practices that are used.

willemstuursma:
Hi @Waterdhavian. We've merged a basis for this in #58.

Maybe you could add support to Invoices for this. See the `\PhpTwinfield\ApiConnectors\BankTransactionApiConnector::sendAll()` to get started.

willemstuursma:
No reply from @Waterdhavian, will not be in 2.0.0

PR #58, opened 2018-01-22
Return stuff from sendAll, add BankTransactionMapper

willemstuursma:


tests/UnitTests/ApiConnectors/BankTransferFaaaTest.php
```diff
@@ -0,0 +1,85 @@
+<?php
```
Thijs-Riezebeek: `BankTransferFaaaTest.php`?

willemstuursma:
> Is this made so complex to prevent parsing the entire return XML into objects when that might not be needed?

Yes @Thijs-Riezebeek, this is indeed the reason. Also this allows you to get Exceptions if there was a partial failure.

PR #59, opened 2018-01-22
Add TwinfieldReference interface

willemstuursma:
Implement on Bank-, Journal-, Purchase- and SalesTransactionLines.


src/BaseTransactionLine.php
```diff
@@ -361,4 +364,68 @@ public function setBaseline(?int $baseline)
...
+            private $office;
+
+            /**
+             * @var int|null
+             */
+            private $lineId;
+
+            public function __construct(BaseTransactionLine $line)
+            {
+                /** @var JournalTransaction|PurchaseTransaction|SalesTransaction $transaction */
```
Thijs-Riezebeek: Are the methods used below available on all of these transaction types?

src/Transactions/BankTransactionLine/Base.php
```diff
@@ -52,6 +56,29 @@
...
+    /**
+     * @return BankTransaction
+     */
+    public function getTransaction(): LinesField
+    {
+        return $this->transaction;
+    }
+
+    /**
+     * @param BankTransaction $object
```
Thijs-Riezebeek: `$object` is a bit generic.

Also, why is this type hinted as as `BankTransaction` in the `@param` and `@return`, but the code type hints it as `LinesField`.

src/PurchaseTransactionLine.php
```diff
@@ -16,6 +18,30 @@ class PurchaseTransactionLine extends BaseTransactionLine
...
 
+    /**
+     * @var PurchaseTransaction
+     */
+    private $transaction;
+
+    /**
+     * @param PurchaseTransaction $object
+     */
+    public function setTransaction($object): void
```
Thijs-Riezebeek: Perhaps the `setTransaction` method should throw an exception if a transaction is already set. Just to prevent weird behavior when adding a line to multiple transactions for whatever reason.

src/Transactions/TransactionLine.php
```diff
@@ -3,9 +3,39 @@
...
+     * @return int|null
+     */
     public function getId(): ?int;
+
+    /**
+     * Set the bank transaction on the line. This is needed later on.
+     *
+     * @param LinesField $object
+     */
+    public function setTransaction($object): void;
```
Thijs-Riezebeek: Perhaps the visitor pattern is more suited for this job?

Instead of calling `TransactionLine::setTransaction()` on a line whenever it is added to a transaction and exposing the entire transaction, you would call `TransactionLine::visit(LinesField $object)`, which would then scan the transaction for any fields that it needs. This interface could then expose the office and so on retrieved from the transaction.

src/MatchReference.php
```diff
@@ -0,0 +1,45 @@
...
+namespace PhpTwinfield;
+
+class MatchReference implements MatchReferenceInterface
+{
+    /**
+     * @var Office
+     */
+    private $office;
+
+    private $number;
```
Thijs-Riezebeek: Missing a number of `@var` annotation

src/MatchSet.php
```diff
@@ -63,10 +64,13 @@ public function setMatchDate(\DateTimeInterface $matchDate)
 
     /**
      * @param MatchLine $line
+     * @internal  Don't call this, use MatchLine::create
```
Thijs-Riezebeek: method name is not accurate anymore

PR #64, opened 2018-01-24
DebitCredit field is wrong for lines of 'total' type, fixes #25

### LINKED ISSUE #25: DebitCredit field is wrong for lines of 'total' type
According to the docs:

> If line type = `total`
> * In case of a 'normal' sales transaction `debit`.
> * In case of a credit sales transaction `credit`.
> 
> If line type = `detail` or `vat`
> * In case of a 'normal' sales transaction `credit`.
> * In case of a credit sales transaction `debit`.

Note that the values are switched for `total` lines.

Currently, when you set the value of a transaction using `setValue()`, the `debitcredit` field is also set. But for total lines it is not set correctly.

Daanvm:
I had to rename `getType()` on transaction lines to `getLineType()` because there already existed a `getType()` method on electronic bank statement transactions with another meaning.

This is done in a separate commit, the real change is done in 1cb690a.

### DIFF
--- a/src/BankTransaction.php
+++ b/src/BankTransaction.php
@@ -107,7 +107,7 @@ class BankTransaction implements Transaction
          */
         $this->traitAddLine($line);
 
-        if (!$line->getType()->equals(LineType::TOTAL())) {
+        if (!$line->getLineType()->equals(LineType::TOTAL())) {
             /*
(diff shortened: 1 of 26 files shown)


Daanvm:
Wait, I now see that the debitcredit field for purchase transactions is completely the opposite of that of sales transactions, which makes sense of course.

I'll make a new PR when I've fixed this.

PR #68, opened 2018-01-25
Fix customers send all

Thijs-Riezebeek:
The `CustomerApiConnector::sendAll()` was broken because `CustomersDocument` has an implementation that only supported once customer.

This resulted in always having 1 `<dimension>` element with duplicate info for all customers (e.g. calling `$document->addCustomer($customer)` 20 times would result in 20 `<code>` elements, etc.


willemstuursma:
:+1:

Issue #69, opened 2018-01-31
Mysteriously logged out during operations


willemstuursma:
We got this error a few times:

```
E_ERROR: Uncaught SoapFault exception: [soap:Client] Your logon credentials are not valid anymore. Try to log on again. in vendor/php-twinfield/twinfield/src/Services/ProcessXmlService.php:50
Stack trace:
#0 vendor/php-twinfield/twinfield/src/Services/ProcessXmlService.php(50): SoapClient->__call('ProcessXmlStrin...', Array)
#1 vendor/php-twinfield/twinfield/src/ApiConnectors/BankTransactionApiConnector.php(56): PhpTwinfield\Services\ProcessXmlService->sendDocument(Object(PhpTwinfield\DomDocuments\BankTransactionDocument))
```

For reliability, we should automate re-connecting.

willemstuursma:
(cc @Thijs-Riezebeek)

Daanvm:
We found out this was caused by us both using the same credentials in our scripts. As soon as @Thijs-Riezebeek started his script, my session was made invalid.

I assume this will no longer be an issue when we switch over to oauth authentication, but you know what they say about assumptions...

willemstuursma:
Ah. I think it is still good to merge these changes, the logged in person will flip all the time but at least it will work.

marceldecroon:
Hi. Switching to oauth will definitely solve this problem. By doing so, you don't need to login in order to open a session. All you have to do is passing the access_token as part of the request to Twinfield.

PR #70, opened 2018-01-31
Recover from logon failures #69

willemstuursma:


willemstuursma:
There's no need to keep copies of the services in `BaseApiConnector`, they are kept in `Connection` already.

willemstuursma:
@Daanvm

PR #72, opened 2018-02-02
Added support for parsing the lines in BankTransactions.

### LINKED ISSUE #71: Returned bank transactions always have 0 lines
The `\PhpTwinfield\Mappers\BankTransactionMapper::map()` method that transforms Twinfield's response to a `BankTransaction` completely ignores any lines, making it not that useful.

mvdpanne:
Some fields could not be parsed, as they are not part of the `BankTransactionLine` classes. Examples of such fields are `currencydate` and `basevalue`.

~~Furthermore, no example XML line for the VAT type is available, so no tests have been added for the parsing of such lines.~~ This has now been fixed, given that a new example XML response containing all line types has been provided by @Daanvm.

This PR fixes #71.

### DIFF
--- a/src/Mappers/BankTransactionMapper.php
+++ b/src/Mappers/BankTransactionMapper.php
@@ -1,33 +1,199 @@
 <?php
 namespace PhpTwinfield\Mappers;
 
+use PhpTwinfield\Enums\PerformanceType;
+use PhpTwinfield\Office;
+use PhpTwinfield\Transactions\BankTransactionLine\Base;
+use PhpTwinfield\Transactions\BankTransactionLine\Detail;
+use PhpTwinfield\Transactions\BankTransactionLine\Total;
+use PhpTwinfield\Transactions\BankTransactionLine\Vat;
 use PhpTwinfield\BankTransaction;
+use PhpTwinfield\Enums\DebitCredit;
+use PhpTwinfield\Enums\LineType;
 use PhpTwinfield\Util;
 
 class BankTransactionMapper extends BaseMapper
 {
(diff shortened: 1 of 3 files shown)


willemstuursma:
> Furthermore, no example XML line for the VAT type is available, so no tests have been added for the parsing of such lines.

@Daanvm can you arrange this?

Daanvm:
> @Daanvm can you arrange this?

I've sent @mvdpanne an example of such a response. He'll add them to the test 👍

src/Mappers/BankTransactionMapper.php
```diff
@@ -144,11 +146,14 @@ private static function setBankTransactionLineBaseFields(
         Base $line
     ): void {
         $line->setId($lineElement->getAttribute("id"));
-        $line->setDescription(self::getField($lineElement, "description"));
         $value = self::getField($lineElement, 'value');
         $line->setValue(Util::parseMoney($value, $bankTransaction->getCurrency()));
         $line->setInvoiceNumber(self::getField($lineElement, "invoicenumber"));
 
+        $description = self::getField($lineElement, "description");
+        if ($description) {
```
willemstuursma: This is always a bit tricky because `"0"` would be a valid description. I recommend explicit `null` checks instead.

willemstuursma:
Great work @mvdpanne, one minor change requested and then go ahead and merge.

src/Mappers/BankTransactionMapper.php
mvdpanne: Do you mean an explicit `null` check for description only, or in general?

PR #74, opened 2018-02-05
Added support for processing all objects when using 'sendAll()'.

### LINKED ISSUE #73: sendAll() should not stop after first error
When sending objects in chunks, `sendAll()` throws an exception as soon as one object in a chunk contains an error. This means that all following chunks aren't even sent to Twinfield.

Secondly, throwing an exception when one object contains an error makes it impossible to get the returned responses of the successful objects.

**Suggested solution**
When we remove the `$response->assertSuccessful()` line from `PhpTwinfield\Services\ProcessXmlService::sendDocument()`, all chucks will be processed and an array of `IndividualMappedResponse` will be returned. The `assertSuccessful()` check is only done once the response is unwrapped.

Instead of returning an array of `IndividualMappedResponse` we should return a response object with methods as `hasSuccessfulResponses()`, `countSuccessfulResponses()`, `getSuccessfulResponses()`, `hasFailedResponses()`, `countFailedResponses()`, `getFailedResponses()`, `getAllResponses()` and `assertAllSuccessful()`.

mvdpanne:
Instead of stopping as soon as an object contains an error, now all objects that are returned by Twinfield will be processed. All `IndividualMappedResponse` objects will furthermore be wrapped into a `MappedResponseCollection` object. This class has methods to easily determine whether the objects returned by Twinfield have errors/are successful, how many of them have errors/are successful and return the `IndividualMappedResponse` objects that have errors/are successful.

This PR fixes #73.

### DIFF
--- a/src/ApiConnectors/BankTransactionApiConnector.php
+++ b/src/ApiConnectors/BankTransactionApiConnector.php
@@ -21,12 +21,12 @@ class BankTransactionApiConnector extends BaseApiConnector
      */
     public function send(BankTransaction $bankTransaction): BankTransaction
     {
-        $result = $this->sendAll([$bankTransaction]);
+        $bankTransactionResponses = $this->sendAll([$bankTransaction]);
 
-        Assert::count($result, 1);
+        Assert::count($bankTransactionResponses, 1);
 
-        foreach ($result as $each) {
-            return $each->unwrap();
+        foreach ($bankTransactionResponses as $bankTransactionResponse) {
+            return $bankTransactionResponse->unwrap();
         }
     }
 
--- a/src/ApiConnectors/BankTransactionApiConnector.php
+++ b/src/ApiConnectors/BankTransactionApiConnector.php
@@ -6,7 +6,7 @@ use PhpTwinfield\BankTransaction;
 use PhpTwinfield\DomDocuments\BankTransactionDocument;
 use PhpTwinfield\Exception;
 use PhpTwinfield\Mappers\BankTransactionMapper;
-use PhpTwinfield\Response\IndividualMappedResponse;
+use PhpTwinfield\Response\MappedResponseCollection;
 use PhpTwinfield\Response\Response;
 use Webmozart\Assert\Assert;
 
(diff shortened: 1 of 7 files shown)


src/Response/MappedResponseCollection.php
```diff
@@ -0,0 +1,93 @@
...
+     * @return IndividualMappedResponse[]
+     */
+    public function getFailedResponses()
+    {
+        return $this->getResponses(false);
+    }
+
+    public function assertAllSuccessful()
+    {
+        assert($this->countResponses(false) === 0);
```
Thijs-Riezebeek: Not sure, but maybe you should use webmozart's `Assert` here. Lets ask Assert-guru @willemstuursma.

src/Response/MappedResponseCollection.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * @var bool $successful
+     *
+     * @return IndividualMappedResponse[]
+     */
+    private function getResponses(bool $successful)
```
Thijs-Riezebeek: `: array` return type is missing

src/ApiConnectors/BankTransactionApiConnector.php
```diff
@@ -21,21 +21,21 @@ class BankTransactionApiConnector extends BaseApiConnector
      */
     public function send(BankTransaction $bankTransaction): BankTransaction
     {
-        $responses = $this->sendAll([$bankTransaction]);
+        $result = $this->sendAll([$bankTransaction]);
```
Thijs-Riezebeek: I'd give this a bit more of a useful variable name. `$bank_transaction_responses`. `foreach ($bank_transaction_responses as $bank_transaction_response)`.

src/Response/MappedResponseCollection.php
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * @return IndividualMappedResponse[]
+     */
+    public function getFailedResponses(): array
+    {
+        return $this->getResponses(false);
+    }
+
+    public function assertAllSuccessful(): void
```
Thijs-Riezebeek: Could you add an `@throws InvalidArgumentException` here, since that is the exception that is thrown when the assertion fails. That way the caller knows how to handle this.

Thijs-Riezebeek: This should actually have its own exception class, since it is not an invalid argument, but it won't matter that much.

Thijs-Riezebeek:
Approved

PR #75, opened 2018-02-07
OAuth2 and OAuth2 Provider implementation

mvdpanne:
This Pull Request adds support for OAuth2 as authentication method for the Twinfield API client.

In order to make this possible, a Twinfield OAuth2 Provider (which extends the base client of `league/oauth2-client`) had to be implemented. This is also featured in the [oauth2-provider](https://github.com/php-twinfield/twinfield/tree/oauth2-provider) branch, but this branch has been merged with the `oauth2` branch and is therefore part of this PR.

In a few quick words (to be elaborated on later, perhaps), the following has been done:
- Remove the original `Config` and `Connection` classes in favor of an abstract `\PhpTwinfield\Secure\AuthenticatedConnection` class with subclasses `\PhpTwinfield\Secure\WebservicesAuthentication` and `\PhpTwinfield\UnitTests\Secure\OpenIdConnectionAuthenticationTest`.
- Added OAuthProvider class.
- Made sure the codebase can use the new methods of logging in.

Left to do:
- ~~Updates to `README.md`~~.
- Tests for `\PhpTwinfield\Secure\WebservicesAuthentication`. Is this necessary? The fact that it uses an anonymous class makes that rather difficult (at least to me).

Remarks:
- Instead of having to pass an `Office` object to the `PhpTwinfield\Secure\OpenIdConnectAuthentication` class, it could be an idea to request the list of `Office`'s from Twinfield, and using the first of that list by default.
- Something that came up during the implementation: it seems that not all `ApiConnectors` pass an `officeCode` in their SOAP bodies. This should be done, because when using OAuth2, an `officeCode` is present in the SOAP header, which is used in case no such code is defined in the SOAP body.
- The `twf.user` scope does not work, and an exception will be thrown if it is used when creating an authorization url in the OAuthProvider. I sent out an inquiry to Twinfield about this.
- Furthermore, the endpoint for retrieving resource owner information does not work. For now, using the method to retrieve such information (The `League\OAuth2\Client\Provider\AbstractProvider::getResourceOwner` method in the OAuthProvider) will throw an exception. Likewise, I also contacted Twinfield about this.

Last but not least, thanks to @alexjeen, @willemstuursma and @Daanvm for their work.


willemstuursma:
Thanks @mvdpanne, I'll review this on Monday.

willemstuursma:
When this is merged, can you open a PR here https://github.com/thephpleague/oauth2-client/blob/master/docs/providers/thirdparty.md

willemstuursma:
Well done, some minor stuff.

PR #78, opened 2018-02-09
Added support for retrying requests that return the SSL error.

### LINKED ISSUE #76: Retry call in case of SSL error
Sometimes we get the following error when trying to call the Twinfield API:

```
Error: Uncaught ErrorException: SoapClient::__doRequest(): SSL: Connection reset by peer in /somepath/vendor/php-twinfield/twinfield/src/Services/ProcessXmlService.php:53
Stack trace:
#0 [internal function]: {closure}(2, 'SoapClient::__d...', '/somepath/vendo...', 53, Array)
#1 [internal function]: SoapClient->__doRequest('<?xml version="...', 'https://account...', 'http://www.twin...', 1, 0)
#2 /somepath/vendor/php-twinfield/twinfield/src/Services/ProcessXmlService.php(53): SoapClient->__call('ProcessXmlStrin...', Array)
#3 /somepath/vendor/php-twinfield/twinfield/src/ApiConnectors/BaseApiConnector.php(44): PhpTwinfield\Services\ProcessXmlService->sendDocument(Object(PhpTwinfield\DomDocuments\BankTransactionDocument))
#4 /somepath/vendor/php-twinfield/twinfield/src/ApiConnectors/BankTransactionApiConnector.php(56): PhpTwinfield\ApiConnectors\BaseApiConnector->sendXmlDocument(Ob in /somepath/vendor/php-twinfield/twinfield/src/Services/ProcessXmlService.php:53
```

When we try the request again, it just completes successfully. It would be nice if we could retry the request in this client.

mvdpanne:
This Pull Request fixes #76 by adding the functionality of automatically retrying request in case of the specified SSL error. It will retry a maximum of three times (defined by a constant), after which it will throw an exception anyway.

Because the amount of error messages that we want to retry automatically might increase in the future, I figured I could best create a constant array with these messages, such that the logic of checking whether the request should be retried can be embedded in a for-loop. Thankfully, PHP 7.1 features catching multiple exception class in one catch block, which prevented the need for duplicating code.

I furthermore modified the `\PhpTwinfield\UnitTests\BaseApiConnectorTest` to work for both the SoapFault and ErrorException exception types, and for all error messages that should be retried, using `@dataProvider`'s. It is also tested that after three times, no more retries will be attempted.

### DIFF
--- a/src/ApiConnectors/BaseApiConnector.php
+++ b/src/ApiConnectors/BaseApiConnector.php
@@ -11,11 +11,6 @@ use PhpTwinfield\Services\ProcessXmlService;
 abstract class BaseApiConnector
 {
     /**
-     * @var Connection
-     */
-    private $connection;
-
-    /**
      * @var string[]
      */
     private const RETRY_REQUEST_EXCEPTION_MESSAGES = [
@@ -23,6 +18,21 @@ abstract class BaseApiConnector
         "SSL: Connection reset by peer",
         "Your logon credentials are not valid anymore. Try to log on again."
     ];
+
+    /**
+     * @var int
+     */
+    private const MAX_RETRIES = 3;
+
+    /**
+     * @var Connection
+     */
+    private $connection;
+
+    /**
+     * @var int
+     */
+    private $numRetries = 0;
 
     public function __construct(Connection $connection)
     {
@@ -49,7 +59,10 @@ abstract class BaseApiConnector
      */
     public function sendXmlDocument(\DOMDocument $document) {
         try {
-            return $this->getProcessXmlService()->sendDocument($document);
+            $response = $this->getProcessXmlService()->sendDocument($document);
+            $this->numRetries = 0;
+
+            return $response;
         } catch (\SoapFault | \ErrorException $exception) {
             /*
              * Always reset the client. There may have been TCP connection issues, network issues,
@@ -59,11 +72,18 @@ abstract class BaseApiConnector
 
             /* For a given set of exception messages, always retry the request. */
             foreach (self::RETRY_REQUEST_EXCEPTION_MESSAGES as $message) {
-                if (stripos($exception->getMessage(), $message) !== false) {
-                    return $this->sendXmlDocument($document);
+                if (stripos($exception->getMessage(), $message) === false) {
+                    continue;
                 }
+                $this->numRetries++;
+
+                if ($this->numRetries > self::MAX_RETRIES) {
+                    break;
+                }
+                return $this->sendXmlDocument($document);
             }
 
+            $this->numRetries = 0;
             throw new Exception($exception->getMessage(), 0, $exception);
         }
     }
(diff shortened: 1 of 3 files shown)


Daanvm:
Can you add a retry for the following exception too?

```
E_ERROR: Uncaught SoapFault: Error Fetching http headers in /somepath/vendor/php-twinfield/twinfield/src/Services/ProcessXmlService.php:53
Stack trace:
#0 [internal function]: SoapClient->__doRequest('<?xml version="...', 'https://account...', 'http://www.twin...', 1, 0)
#1 /somepath/vendor/php-twinfield/twinfield/src/Services/ProcessXmlService.php(53): SoapClient->__call('ProcessXmlStrin...', Array)
#2 /somepath/vendor/php-twinfield/twinfield/src/ApiConnectors/BaseApiConnector.php(44): PhpTwinfield\Services\ProcessXmlService->sendDocument(Object(PhpTwinfield\DomDocuments\CustomersDocument))
#3 /somepath/vendor/php-twinfield/twinfield/src/ApiConnectors/CustomerApiConnector.php(126): PhpTwinfield\ApiConnectors\BaseApiConnector->sendXmlDocument(Object(PhpTwinfield\DomDocuments\CustomersDocument))
```

src/ApiConnectors/BaseApiConnector.php
```diff
@@ -42,21 +50,21 @@ protected function getProcessXmlService(): ProcessXmlService
...
 
-            if (stripos($soapFault->getMessage(), "Your logon credentials are not valid anymore. Try to log on again.") !== false) {
-                /*
-                 * Automatically retry and log on again.
-                 */
-                return $this->sendXmlDocument($document);
+            /* For a given set of exception messages, always retry the request. */
+            foreach (self::RETRY_REQUEST_EXCEPTION_MESSAGES as $message) {
+                if (stripos($exception->getMessage(), $message) !== false) {
+                    return $this->sendXmlDocument($document);
```
Daanvm: This will create an infinite loop with infinite API calls if the error will be returned consistently. Can you change this to stop trying after 3 tries?

Daanvm:
LGTM!

src/ApiConnectors/BaseApiConnector.php
```diff
@@ -5,17 +5,35 @@
...
 use PhpTwinfield\Services\FinderService;
 use PhpTwinfield\Services\ProcessXmlService;
 
 abstract class BaseApiConnector
 {
+    /**
+     * @var string[]
+     */
+    private const RETRY_REQUEST_EXCEPTION_MESSAGES = [
+        "Error Fetching http headers",
```
willemstuursma: @mvdpanne I am not sure if the `Error Fetching http headers` should be retried, it may have been received (and processed) by the server. I think you get this during a timeout.

mvdpanne: Hmm, I looked it up and you're right. It means the connection has been made, but the response does not return in time. So I guess you could always say that the server has processed/is still processing the request.

So yeah, seems right to remove it. What do you think @Daanvm?

willemstuursma: I'll take care of it, thanks!

Daanvm: Yeah, you're right. I'm sorry.

PR #79, opened 2018-02-12
Require only a refresh token for OAuth2 authentication

mvdpanne:
This PR improves the implementation of OAuth2 authentication.

Previously, an `\League\OAuth2\Client\Token\AccessToken` was expected when creating an instance of the `\PhpTwinfield\Secure\OpenIdConnectAuthentication` class. This `AccessToken` has to contain an access token.

The problem is that only providing a refresh token is enough, as a valid access token can be retrieved using the refresh token. This has now been implemented in the code.

### DIFF
--- a/src/Secure/OpenIdConnectAuthentication.php
+++ b/src/Secure/OpenIdConnectAuthentication.php
@@ -2,7 +2,6 @@
 
 namespace PhpTwinfield\Secure;
 
-use League\OAuth2\Client\Token\AccessToken;
 use PhpTwinfield\Office;
 use PhpTwinfield\Secure\Provider\InvalidAccessTokenException;
 use PhpTwinfield\Secure\Provider\OAuthProvider;
@@ -21,9 +20,14 @@ class OpenIdConnectAuthentication extends AuthenticatedConnection
     private $provider;
 
     /**
-     * @var AccessToken
+     * @var null|string
      */
     private $accessToken;
+
+    /**
+     * @var string
+     */
+    private $refreshToken;
 
     /**
      * @var Office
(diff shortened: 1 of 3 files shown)


src/Secure/OpenIdConnectAuthentication.php
```diff
@@ -35,24 +39,17 @@ class OpenIdConnectAuthentication extends AuthenticatedConnection
...
     {
-        $this->provider    = $provider;
-        $this->accessToken = $accessToken;
+        $this->provider     = $provider;
+        $this->refreshToken = $refreshToken;
 
         /*
          * The office code that is part of the Office object that is passed here will be
          * the default office code used during requests. If an office code is included in
          * the SOAP request body, this will always take precedence over this default.
```
willemstuursma: This comment should be in the phpdoc.

src/Secure/OpenIdConnectAuthentication.php
```diff
@@ -35,24 +39,17 @@ class OpenIdConnectAuthentication extends AuthenticatedConnection
      */
     private $cluster;
 
-    /**
-     * @throws OAuthException
-     */
-    public function __construct(OAuthProvider $provider, AccessToken $accessToken, Office $office)
+    public function __construct(OAuthProvider $provider, string $refreshToken, Office $office)
```
willemstuursma: Please update the readme.

src/Secure/OpenIdConnectAuthentication.php
```diff
@@ -35,24 +39,17 @@ class OpenIdConnectAuthentication extends AuthenticatedConnection
...
     {
-        $this->provider    = $provider;
-        $this->accessToken = $accessToken;
+        $this->provider     = $provider;
+        $this->refreshToken = $refreshToken;
 
         /*
          * The office code that is part of the Office object that is passed here will be
```
willemstuursma: This comment should be in the phpdoc not in the method body.

PR #80, opened 2018-02-13
Fix for the 'twf.user' scope and 'getResourceOwner' method

mvdpanne:
After some communication with Twinfield, it became apparent that the `twf.user` scope was not working, because the `openid` scope was missing. Unfortunately, this was not stated in their documentation (which will be updated). Furthermore, the problem with the `getResourceOwner` method is fixed. It can be called only when the `openid` scope is used. The `twf.user` scope can optionally be passed to retrieve more information from the `getResourceOwner` method.

Changes:
- Added the `openid` scope as constant.
- Added the `openid` scope as a default scope, as the default scopes should include those scopes that are necessary to request the details of the resource owner.
- Made sure to allow the `twf.user` scope again, but only if the `openid` scope is present.
- Modified the `ResourceOwner` class to match the data that Twinfield returns.

### DIFF
--- a/src/Secure/Provider/OAuthProvider.php
+++ b/src/Secure/Provider/OAuthProvider.php
@@ -17,8 +17,7 @@ class OAuthProvider extends AbstractProvider
     use BearerAuthorizationTrait;
 
     /**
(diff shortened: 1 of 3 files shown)


willemstuursma:
Cheers @mvdpanne

Issue #77, opened 2018-02-09
Error while sending invoice


CasNelissen:
When we're trying to send an invoice, we get the following error:

"The element `unitspriceexcl` can not be supplied"

We've got the following code:

`
$customer = new Customer($config->getConfig());
$customer->setOffice(9999);

$customer = $customer->getCustomer("1001");

$invoice = new Invoice($config->getConfig());
$invoice->setOffice(9999);

$line = new InvoiceLine();
$line
    ->setQuantity(1.00)
    ->setArticle(123456789)
    ->setUnits(1)
    ->setVatCode("01SHT")
    ->setDim1(800500)
    ->setDescription("TEST")
    ->setAllowDiscountOrPremium(false)
;

$twinfieldInvoice = new TwinfieldInvoice();
$twinfieldInvoice
    ->setCustomer($customer)
    ->addLine($line)
    ->setStatus("concept")
    ->setPaymentMethod("bank")
    ->setInvoiceType("FACTUUR")
    ->setInvoiceDate((new DateTime())->format('Ymd'))
    ->setCurrency("EUR")
    ->setBank("BNK")
    ->setPeriod("2018/02")
    ->setDueDate((new DateTime())->modify("+30 days"))
;
`

Maybe it's me, but I don't see any `setUnitsPriceExcl` function call.

Daanvm:
Hi @CasNelissen,
There should probably be a check to see if `unitspriceexcl` is empty, and if it is it should not be added to the XML that is sent to Twinfield. There is currently no such check in the `InvoicesDocument` class.

If you could create a pull request, I am more than willing to merge it!

PR #81, opened 2018-02-13
Fix for 'Error while sending invoice #77' ticket.

CasNelissen:
Fix for 'Error while sending invoice #77' ticket. Plus changed isset to !empty.


willemstuursma:
@CasNelissen This looks good to me, could you update the `InvoiceIntegrationTest` which is now failing?

willemstuursma:
That's great @CasNelissen, but it would be even better with the tags gone and not commented out.

PR #83, opened 2018-02-19
Fix 'getWriteOffType' return type

### LINKED ISSUE #82: `\PhpTwinfield\MatchLine::getWriteOffType()` has incorrect return type
`\PhpTwinfield\MatchLine::getWriteOffType()` should have a nullable return type, since there is no guarantee that this field is set.

This currently results in a `TypeError` when calling this method when the value is not set.

mvdpanne:
As the title states, this PR fixes the `\PhpTwinfield\MatchLine::getWriteOffType` return type.

Furthermore, removed some unnecessary/redundant information from docblocks. No change is required in `\PhpTwinfield\DomDocuments\MatchDocument::createLineElement` (where the method is being used), because when `\PhpTwinfield\MatchLine::$writeoff` is non-null, it is guaranteed that `\PhpTwinfield\MatchLine::$writeofftype` is non-null.

This PR fixes #82.

### DIFF
--- a/src/MatchLine.php
+++ b/src/MatchLine.php
@@ -34,24 +34,19 @@ class MatchLine
     private $matchvalue;
 
     /**
-     * @var Money|NULL
+     * @var Money|null
      */
     private $writeoff;
 
     /**
-     * @var Enums\WriteOffType
+     * @var Enums\WriteOffType|null
      */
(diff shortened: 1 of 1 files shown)


willemstuursma:
LGTM.

PR #90, opened 2018-02-20
Add `BookingReferenceInterface` for identifying bookings

Thijs-Riezebeek:
Deleting transactions, statements, etc. is done using a reference to
that object. However, the existing `MatchReferenceInterface` does not
suffice for this, since it references a specific line within such an
object.

Therefore `BookingReferenceInterface` was added to unqiuely identify a
booking.

`BookingReferenceInterface` is not extended by `MatchReferenceInterface`
on purpose, since we don't want anyone to accidentally pass a
`MatchReferenceInterface` to a delete method, expecting the specific
line to be deleted, but it actually deletes the entire transaction.
Usage of `BookingReferenceInterface` should be explicit.

Ease of use methods `BookingReference::fromMatchReference()` and
`MatchReference::fromBookingReference()` have been added.

### DIFF
--- a/src/BookingReferenceInterface.php
+++ b/src/BookingReferenceInterface.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace PhpTwinfield;
+
+/**
+ * The values provided by this interface uniquely define a booking (e.g. SalesTransaction, BankTransaction, etc.) in
+ * Twinfield and can be used when a reference to the booking is needed, for example when deleting a booking.
+ *
+ * You can use the provided MatchReference object or you can implement this directly on one of your entities / domain
+ * models.
+ */
+interface BookingReferenceInterface
+{
+    /**
+     * References the office in which the transaction was booked.
+     */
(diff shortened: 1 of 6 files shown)


willemstuursma:
> since we don't want anyone to accidentally pass a
MatchReferenceInterface to a delete method. 

That's very smart

src/BookingReferenceInterface.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace PhpTwinfield;
+
+/**
+ * The values provided by this interface uniquely define a booking (e.g. SalesTransaction, BankTransaction, etc.) in
+ * Twinfield and can be used when a reference to the booking is needed, for example when deleting a booking.
+ *
+ * You can use the provided MatchReference object or you can implement this directly on one of your entities / domain
```
Thijs-Riezebeek: Oops

willemstuursma:
LGTM

PR #92, opened 2018-02-23
89 delete banktransactions

### LINKED ISSUE #89: Delete provisional transactions
It is possible to delete provisional transactions with the following code:

```php
$document = new DOMDocument();
$document->loadXML(<<<XML
<transaction action="delete" reason="Test transaction">
    <office>OFFICE001</office>
    <code>BNK</code>
    <number>201700006</number>
</transaction>
XML
);
$bankTransactionApiConnector->sendXmlDocument($document);
```

I would like to see a `delete()` method directly on both the `TransactionApiConnector` and the `BankTransactionApiConnector`. Note that the reason is required, it will fail if the reason is not included.

### Result
Verification of the result works the same as when adding a new transaction, with `result` and `msg` attribute on the root element:

```xml
<?xml version="1.0"?>
<transaction action="delete" reason="Test transaction" msgtype="error" msg="De status van de boeking moet Concept zijn." result="0">
    <office>OFFICE001</office>
    <code>BNK</code>
    <number>201700006</number>
</transaction>
```

willemstuursma:
This fixes #89.


src/Transactions/TransactionFields/CodeNumberOfficeFields.php
```diff
@@ -64,4 +65,18 @@ public function getNumber(): ?int
     {
         return $this->number;
     }
+
+    /**
+     * Get the booking reference. The booking reference uniquely identifies a booking.
+     *
+     * @return BookingReference
+     */
+    public function getBookingReference(): BookingReference
```
Thijs-Riezebeek: I believe that the method on all of the `Line` objects is called `getReference(): MatchReferenceInterface`. So perhaps we should call this `getReference()` as well?

tests/UnitTests/ApiConnectors/BankTransactionApiConnectorTest.php
```diff
@@ -195,4 +196,33 @@ public function testSendAllReturnsMappedObjectsAllLineTypes()
         $this->assertEquals(Money::EUR(10000), $line->getVatTurnover());
         $this->assertEquals(Money::EUR(10000), $line->getVatBaseTurnover());
     }
+
+    /**
+     * @expectedException Exception
+     * @expectedExceptionMessage De status van de boeking moet Concept zijn
+     */
+    public function testDeletionWorks()
```
Thijs-Riezebeek: This looks more like `testDeleteThrowsWhenReturnedXmlContainsErrors`.

Can we also test the happy path?

Thijs-Riezebeek: This sounds more like `testDeleteThrowsWhenResponseContainsErrorMessages`.

Can we test the happy path as well?

src/Transactions/TransactionFields/CodeNumberOfficeFields.php
Thijs-Riezebeek: I believe that the method on all of the `Line` objects is called `getReference(): MatchReferenceInterface`. So perhaps we should call this `getReference()` as well?

tests/UnitTests/ApiConnectors/BankTransactionApiConnectorTest.php
willemstuursma: I don't think it adds to much value, there's no special code path for happy path. 

If you insist, you can provide me with a successful response and I'll create a test around it.

PR #94, opened 2018-03-16
Add psr/log and have `BaseApiConnector` log info, debug and error logs

Thijs-Riezebeek:
I've added logging to the `BaseApiConnector` so that it becomes possible to debug the actual requests and responses send and received from Twinfield.

### DIFF
--- a/src/ApiConnectors/BaseApiConnector.php
+++ b/src/ApiConnectors/BaseApiConnector.php
@@ -11,7 +11,6 @@ use PhpTwinfield\Services\ProcessXmlService;
 use PhpTwinfield\Util;
 use Psr\Log\LoggerAwareInterface;
 use Psr\Log\LoggerAwareTrait;
-use Psr\Log\LoggerTrait;
 
 abstract class BaseApiConnector implements LoggerAwareInterface
 {
--- a/src/ApiConnectors/BaseApiConnector.php
+++ b/src/ApiConnectors/BaseApiConnector.php
@@ -109,10 +109,10 @@ abstract class BaseApiConnector implements LoggerAwareInterface
             return;
         }
 
-        $message = sprintf(
-            "Sending request to Twinfield.%s",
-            $this->numRetries > 0 ? ' (attempt ' . ($this->numRetries + 1) . ')' : ''
-        );
+        $message = "Sending request to Twinfield.";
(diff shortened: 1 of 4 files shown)


tests/UnitTests/ApiConnectors/BaseApiConnectorTest.php
```diff
@@ -73,6 +83,155 @@ private function getMaxNumRetries(): int
...
+                $this->throwException(new \SoapFault('Server', 'Internal server error'))
+            ));
+
+        $this->service->setLogger($this);
+        try {
+            $this->service->sendXmlDocument($request_document);
+        } catch (\PhpTwinfield\Exception $e) {
+        }
+
+        self::assertCount(2, $this->logs);
```
mvdpanne: Wondering why only two logs? If it fails it should have first retried three times, right?

src/ApiConnectors/BaseApiConnector.php
```diff
@@ -82,14 +92,68 @@ public function sendXmlDocument(\DOMDocument $document) {
...
         }
     }
 
+    private function logSendingDocument(\DOMDocument $document): void
+    {
+        if (!$this->logger) {
+            return;
+        }
+
+        $message = sprintf(
```
mvdpanne: I think the following would be slightly more clear:
```
$message = 'Sending request to Twinfield.';
if ($this->numRetries > 0) {
    $message .= ' (attempt ' . ($this->numRetries + 1) . ')';
}
```

Take whatever feels best to you.

tests/UnitTests/ApiConnectors/BaseApiConnectorTest.php
Thijs-Riezebeek: This test actually mocks the ProcessXmlService to not fail, but just return a response on the first request. So no need for retries cause we're testing a successful request here :) So there's 1 request and 1 response, so 2 logs.

mvdpanne: Aha, got it 😉

src/ApiConnectors/BaseApiConnector.php
Thijs-Riezebeek: Changed

PR #96, opened 2018-03-19
ArticleMapper Extend BaseMapper

CasNelissen:
Changed the ArticleMapper to extend the BaseMapper and use setFromTagValue to handle all tags. This because functions ('setOffice') failed on type.


willemstuursma:
Thanks, great addition!

PR #97, opened 2018-03-20
Thijs/keep line order consistent

Thijs-Riezebeek:
* Lines that are added to a `BankTranasction` (or any other object using the `LineField` trait) will now be retrieved in the same order as they were added. Note that they will still be ordered by line type as the primary order (TOTAL, then DETAIL, then VAT).

### DIFF
--- a/src/Transactions/TransactionFields/LinesField.php
+++ b/src/Transactions/TransactionFields/LinesField.php
@@ -22,6 +22,13 @@ trait LinesField
      */
     abstract public function getLineClassName(): string;
 
+    protected function getLineCount(): int
+    {
+        return $this->getLineCountForType(LineType::TOTAL())
+            + $this->getLineCountForType(LineType::DETAIL())
+            + $this->getLineCountForType(LineType::VAT());
+    }
+
     private function getLineCountForType(LineType $line_type): int
     {
         if (\array_key_exists($line_type->getValue(), $this->lines_per_type)) {
--- a/src/Transactions/TransactionFields/LinesField.php
+++ b/src/Transactions/TransactionFields/LinesField.php
@@ -14,7 +14,11 @@ trait LinesField
     /**
      * @var TransactionLine[][]
      */
-    private $lines_per_type = [];
+    private $lines_per_type = [
+        LineType::TOTAL  => [],
+        LineType::DETAIL => [],
+        LineType::VAT    => [],
+    ];
 
     /**
      * @return string The class name for transaction lines supported by this transaction. Must be an implementation of
@@ -31,11 +35,7 @@ trait LinesField
 
     private function getLineCountForType(LineType $line_type): int
     {
-        if (\array_key_exists($line_type->getValue(), $this->lines_per_type)) {
-            return count($this->lines_per_type[$line_type->getValue()]);
-        }
-
-        return 0;
+        return count($this->lines_per_type[$line_type->getValue()]);
     }
 
     /**
@@ -48,45 +48,26 @@ trait LinesField
          * one or more detail lines and optionally one or more vat lines. Twinfield returns an error when the lines are
          * in an incorrect order.
          */
-        $lines = [];
-
-        foreach ($this->getTotalLines() as $line) {
-            $lines[] = $line;
-        }
-
-        foreach ($this->getDetailLines() as $line) {
-            $lines[] = $line;
-        }
-
-        foreach ($this->getVatLines() as $line) {
-            $lines[] = $line;
-        }
-
-        return $lines;
+        return \array_merge(
+            $this->getTotalLines(),
+            $this->getDetailLines(),
+            $this->getVatLines()
+        );
     }
 
     private function getTotalLines(): array
     {
-        return $this->getLinesOfType(LineType::TOTAL());
-    }
-
-    private function getLinesOfType(LineType $line_type): array
-    {
-        if (\array_key_exists($line_type->getValue(), $this->lines_per_type)) {
-            return $this->lines_per_type[$line_type->getValue()];
-        }
-
-        return [];
+        return $this->lines_per_type[LineType::TOTAL()->getValue()];
     }
 
     private function getDetailLines(): array
     {
-        return $this->getLinesOfType(LineType::DETAIL());
+        return $this->lines_per_type[LineType::DETAIL()->getValue()];
     }
 
     private function getVatLines(): array
     {
-        return $this->getLinesOfType(LineType::VAT());
+        return $this->lines_per_type[LineType::VAT()->getValue()];
     }
 
     /**
@@ -97,21 +78,14 @@ trait LinesField
     {
         Assert::isInstanceOf($line, $this->getLineClassName());
 
-        $this->addLinePerType($line);
+        if ($line->getLineType()->equals(LineType::TOTAL())) {
+            Assert::eq($this->getLineCountForType(LineType::TOTAL()), 0, 'Cannot set a second TOTAL line');
+        }
+
+        $this->lines_per_type[$line->getLineType()->getValue()][] = $line;
         $line->setTransaction($this);
 
         return $this;
-    }
-
-    private function addLinePerType(TransactionLine $line)
-    {
-        $line_type = $line->getLineType()->getValue();
-
-        if (!\array_key_exists($line_type, $this->lines_per_type)) {
-            $this->lines_per_type[$line_type] = [];
-        }
-
-        $this->lines_per_type[$line_type][] = $line;
     }
 
     /**
--- a/src/Transactions/TransactionFields/LinesField.php
+++ b/src/Transactions/TransactionFields/LinesField.php
@@ -12,15 +12,24 @@ use Webmozart\Assert\Assert;
 trait LinesField
 {
     /**
-     * @var TransactionLine[]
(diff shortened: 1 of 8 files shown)


src/Transactions/TransactionFields/LinesField.php
```diff
@@ -12,16 +12,32 @@
 trait LinesField
 {
     /**
-     * @var TransactionLine[]
+     * @var TransactionLine[][]
      */
-    private $lines = [];
+    private $lines_per_type = [];
```
Daanvm: If you make the `LineType` constants public, you can initialise this array with empty lists. It would not only make the structure of the contents of this variable clear, it would also allow you to remove all `array_key_exists()` checks in this trait.

```php
private $lines_per_type = [
    LineType::TOTAL  => [],
    LineType::DETAIL => [],
    LineType::VAT    => [],
];

File: src/Transactions/TransactionFields/LinesField.php [hunk 2]
```diff
@@ -32,23 +48,45 @@ public function getLines(): array
...
-                if ($type_index_a != $type_index_b) {
-                    return $type_index_a <=> $type_index_b;
-                }
-
-                return $a->getId() <=> $b->getId();
-            }
-        );
-
-        return $this->lines;
+        $lines = [];
```
Daanvm:
```php
return array_merge(
    $this->getTotalLines(),
    $this->getDetailLines(),
    $this->getVatLines()
);
```

src/Transactions/TransactionFields/LinesField.php
```diff
@@ -59,20 +97,30 @@ public function addLine(TransactionLine $line)
     {
         Assert::isInstanceOf($line, $this->getLineClassName());
 
-        $this->lines[] = $line;
-
+        $this->addLinePerType($line);
```
Daanvm: I think the code is more readable if you would just include the code here, instead of creating a new method for it.

tests/UnitTests/BankTransactionUnitTest.php
```diff
@@ -30,4 +34,131 @@ public function testCanGetReferenceFromLine()
         $this->assertEquals("1", $reference->getLineId());
         $this->assertEquals("MEMO", $reference->getCode());
     }
+
+    public function testGetLineReturnsLinesInSameOrderButAlwaysWithTotalAtTheTop()
+    {
+        $total_line_count = 287;
```
Daanvm: Isn't this a bit overkill?

tests/UnitTests/BankTransactionUnitTest.php
```diff
@@ -30,4 +34,131 @@ public function testCanGetReferenceFromLine()
...
+        $bank_transaction = new BankTransaction();
+
+        $this->addDetailLineWithDescription($bank_transaction, '0');
+        $this->addDetailLineWithDescription($bank_transaction, '1');
+        $this->addDetailLineWithDescription($bank_transaction, '2');
+
+        self::assertLinesInOrderBasedOnDescription($bank_transaction, 3);
+    }
+
+    public function testGetLinesReturnsLinesInSameOrderWhenOnlyTotalLineIsAdded()
```
Daanvm: This test doesn't test anything. A list of 1 item can't change order.

tests/UnitTests/BankTransactionUnitTest.php
```diff
@@ -30,4 +34,131 @@ public function testCanGetReferenceFromLine()
...
+    {
+        $bank_transaction = new BankTransaction();
+
+        $this->addTotalLineWithDescription($bank_transaction, '0');
+
+        self::assertLinesInOrderBasedOnDescription($bank_transaction, 1);
+    }
+
+    public function testAdding1TotalLineAnd500DetailLinesToABankTransactionWorks()
```
Daanvm: I'm not sure if it's a good idea to test that 501 items is still OK, the docs explicitly mention the limit of 500, so 501 is a bug, not a feature.

tests/UnitTests/BankTransactionUnitTest.php
```diff
@@ -30,4 +34,131 @@ public function testCanGetReferenceFromLine()
...
+        $bank_transaction = new BankTransaction();
+
+        $this->addTotalLineWithDescription($bank_transaction, '0');
+
+        $this->expectException(\InvalidArgumentException::class);
+
+        for ($i = 1; $i < $total_line_count; $i++) {
+            $this->addDetailLineWithDescription($bank_transaction, (string)$i);
+        }
+    }
```
Daanvm: Can you also add some tests to ensure that VAT lines are at the end of the list?

src/Transactions/TransactionFields/LinesField.php
```diff
@@ -59,20 +97,30 @@ public function addLine(TransactionLine $line)
     {
         Assert::isInstanceOf($line, $this->getLineClassName());
```
Daanvm: With your change you can now also easily ensure that there is at most 1 total line. Multiple total lines are not accepted by Twinfield.

src/Transactions/TransactionFields/LinesField.php
Thijs-Riezebeek: Ahah. Just re-read the docs of `array_merge()` and that would indeed work! I figured it wouldn't work because it overrides values with the same key, but it only does that for string keys, not for integer keys!

src/Transactions/TransactionFields/LinesField.php
Thijs-Riezebeek: Good point! I wanted to do this but couldn't because I couldn't actually call `LineType::TOTAL()`, but it can be done like this. I guess this is one of the reasons to prefer having the constants of the enums publicly available.

src/Transactions/TransactionFields/LinesField.php
Thijs-Riezebeek: Now that the `array_key_exists()` checks are gone, this is just 1 line and I've moved it to this method again.

tests/UnitTests/BankTransactionUnitTest.php
Thijs-Riezebeek: I did a little TDD to find out if this was actually what was causing the random order. So I wrote the test first and it didn't fail when I had `$total_line_count = 10`, only when I went up a bit higher did it start failing. So I figured I'd just do a whole lot. Perhaps it's a little overkill, but luckily unit tests are very fast.

tests/UnitTests/BankTransactionUnitTest.php
Thijs-Riezebeek: The test does test something, just not the order, you're right. I'll rename the test.

tests/UnitTests/BankTransactionUnitTest.php
Thijs-Riezebeek: I will add a similar test to the `SalesTransaction` so that the entire `LineField` trait is covered.

src/Transactions/TransactionFields/LinesField.php
Thijs-Riezebeek: This has really simplified the class, being able to lose the additional methods as well, nice :)

src/Transactions/TransactionFields/LinesField.php
Thijs-Riezebeek: Will add an assertion in `LinesField::addLine()` so that only 1 TOTAL line can be set.

tests/UnitTests/BankTransactionUnitTest.php
Thijs-Riezebeek: Actually agree with this. I've actually changed the implementation to now only allow 500 lines (e.g. 1 TOTAL and 499 DETAIL) and changed the tests to verify this.

src/Transactions/TransactionFields/LinesField.php
Thijs-Riezebeek: ✓

tests/UnitTests/BankTransactionUnitTest.php
```diff
@@ -30,4 +34,147 @@ public function testCanGetReferenceFromLine()
...
+    {
+        $bank_transaction = new BankTransaction();
+
+        $this->addTotalLineWithDescription($bank_transaction, '0');
+
+        try {
+            $this->addTotalLineWithDescription($bank_transaction, '0');
+
+            self::fail('An exception should have been thrown when adding a second total line');
+        } catch (\InvalidArgumentException $e) {
```
Daanvm: I thought you could just call `$this->expectException()` between the first and the second call to `addTotalLineWithDescription()`, so you don't have to do the catching yourself.

tests/UnitTests/BankTransactionUnitTest.php
Thijs-Riezebeek: ✓

tests/UnitTests/BankTransactionUnitTest.php
Thijs-Riezebeek: I was going to do that, but then I noticed a flaw in my implementation.

My implementation would still add the line, and then throw because there are 2 TOTAL lines set on the bank transaction. Obviously, even when throwing, we shouldn't actually set the second TOTAL transaction.

Thus I'm catching the exception here and asserting that there is still only 1 transaction.

PR #100, opened 2018-04-30
Updated Supplier to be more like Customer

manuelderuiter:
- Changed SupplierMapper to extend BaseMapper and use setFromTagValue
- Removed a typo when removing addresses.

Small stuff.


willemstuursma:
Thanks @manuelderuiter. It's a small improvement and there is room for lots more. We're not really using Suppliers a lot at Mollie so any contributions there are welcome. If you see any more room for improvements feel free to open more pull requests.

PR #101, opened 2018-05-08
Added support for Free character field

RickJeroen:
If line type is total and filled with N the sales invoice is excluded from direct debit runs done in Twinfield.

### DIFF
--- a/src/Transactions/TransactionLineFields/FreeCharField.php
+++ b/src/Transactions/TransactionLineFields/FreeCharField.php
@@ -2,6 +2,8 @@
 
 namespace PhpTwinfield\Transactions\TransactionLineFields;
 
+use Webmozart\Assert\Assert;
+
 trait FreeCharField
 {
     /**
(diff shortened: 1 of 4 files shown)


src/Transactions/TransactionLineFields/FreeCharField.php
```diff
@@ -0,0 +1,32 @@
...
+        return $this->freeChar;
+    }
+
+    /**
+     * @param string|null $freeChar
+     * @return $this
+     */
+    public function setFreeChar(?string $freeChar): self
+    {
+        $this->freeChar = $freeChar;
```
willemstuursma: Thanks. Maybe add an `Assert::length($freeChar, 1)` here.

willemstuursma:
Could you also use the new trait in the `BankTransactionLine` class and its friends `\PhpTwinfield\DomDocuments\BankTransactionDocument` and `\PhpTwinfield\Mappers\BankTransactionMapper`?

willemstuursma:
Thanks for opening this pull request!

src/Transactions/TransactionLineFields/FreeCharField.php
RickJeroen: Good one! Done.

PR #103, opened 2018-05-22
Implemented invoice matching and storing financials

Wouter0100:
This PR implements matching invoices with other transactions (using the `MatchReferenceInterface`) and be able to access the financials from invoice.

No BC breaking changes and tested.


src/Mappers/InvoiceMapper.php
```diff
@@ -111,6 +111,35 @@ public static function map(Response $response)
...
+        if ($financialElement !== null) {
+
+            // Go through each financial element and add to the assigned method
+            foreach ($financialsTags as $tag => $method) {
+
+                // Get the dom element
+                $_tag = $financialElement->getElementsByTagName($tag)->item(0);
+
+                // If it has a value, set it to the associated method
+                if (isset($_tag) && isset($_tag->textContent)) {
```
willemstuursma: Could you try to use the methods in `\PhpTwinfield\Mappers\BaseMapper`?

tests/IntegrationTests/InvoiceIntegrationTest.php
```diff
@@ -83,6 +84,64 @@ public function testGetInvoiceWorks()
...
+        $this->assertSame('CoalesceFunctioningOnImpatienceTShirt', $invoiceLine->getDescription());
+        $this->assertSame('15.00', $invoiceLine->getValueExcl());
+        $this->assertSame('0.00', $invoiceLine->getVatValue());
+        $this->assertSame('15.00', $invoiceLine->getValueInc());
+        $this->assertSame('15.00', $invoiceLine->getUnitsPriceExcl());
+        $this->assertSame('8020', $invoiceLine->getDim1());
+
+        // TODO - Vat lines
+
+        $this->assertSame('15.00', $invoice->getTotals()->getValueInc());
```
willemstuursma: Would you mind converting these to `Money\Money` objects, if they cannot have fractional cents?

willemstuursma:
Thanks @Wouter0100 for this very complete contribution!

src/Invoice.php
```diff
@@ -219,4 +222,58 @@ public function setFooterText($footerText)
...
+    public function setFinancialNumber($financialNumber)
+    {
+        $this->financialNumber = $financialNumber;
+	    return $this;
+    }
+
+    /**
+     * References the office in which the transaction was booked.
+     */
+    public function getOffice(): Office
```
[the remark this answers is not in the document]
Wouter0100: Yes - it is, but the return type from the `OfficeField` trait is incompatible with the `MatchReferenceInterface`. The trait makes it nullable, while the interface does not.

src/Invoice.php
```diff
@@ -219,4 +222,58 @@ public function setFooterText($footerText)
...
+    {
+        return $this->financialNumber;
+    }
+
+    /**
+     * Reference the exact line in the transaction.
+     */
+    public function getLineId(): int
+    {
+        return 1;
```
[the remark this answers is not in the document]
Wouter0100: Unfortunately, the 2nd argument in the `MatchLine::addToMatchSet`-method is typehinted to the `MatchReferenceInterface` interface, which `BookingReferenceInterface` isn't and does not inherit in some way.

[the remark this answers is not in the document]
willemstuursma: @Thijs-Riezebeek can we change / tweak this?

tests/IntegrationTests/InvoiceIntegrationTest.php
Wouter0100: Hmm, as far as I can see `$invoice->getTotals()->getValueInc()` currently returns strings, the same as in the `testGetConceptInvoiceWorks`-test. Or how do you mean? :)

src/Mappers/InvoiceMapper.php
Wouter0100: I copied those from the other `Mappers` that have the `financials` implemented, but I've fixed it in this one :+1:.

tests/IntegrationTests/InvoiceIntegrationTest.php
willemstuursma: I would like to change it from returning string to returning `Money` objects, this would indeed break backwards compatibility. 

We're trying to move this library (slowly) to use `Money` Value Objects instead of strings.

Wouter0100: That is okay, or shall I do that in another PR? 

This so you're able to release this PR without BC changes and the other one in the next BC-breaking release? It also seems not really related to this PR. In that PR I'll also be able to update the other places where `financials` are used to use the `BaseMapper`-methods.

src/Invoice.php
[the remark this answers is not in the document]
willemstuursma: If this works for you @Wouter0100 then I guess it's okay. Looking at the Twinfield API docs the transaction line number does not seem optional.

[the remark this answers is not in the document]
Thijs-Riezebeek: @Wouter0100 @willemstuursma You cannot match with a `BookingReferenceInterface`, since you have to match specific lines in the Twinfield API. The `BookingReferenceInterface` is only used as a reference to an entire booking (e.g. a BankTransaction, instead of a specific line in that BankTransaction).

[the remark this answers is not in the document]
Thijs-Riezebeek: When matching invoices, you always want to match the TOTAL line of the invoice. Twinfield wants the TOTAL line to always be at the top of the `<lines>`, so the TOTAL line should always have line ID 1, so I think that this will work, although it does feel a bit hacky.

I'm not too familiar with the `Invoice` class in this package, though, as we've only used the `SalesTransaction` for booking invoices.

src/Invoice.php
```diff
@@ -26,7 +26,7 @@
  * @todo Add documentation and typehints to all properties.
  * @todo Add support for VatLines.
  */
-class Invoice
+class Invoice implements MatchReferenceInterface
```
Thijs-Riezebeek: Perhaps it would be better to have a method `getMatchReference(): MatchReferenceInterface` on the `Invoice` class instead of implementing the interface directly. This way the `Invoice` class doesn't get certain public methods that don't make a lot of sense out of the `MatchReferenceInterface` context, for example, `$invoice->getLineId()` does not make a whole lot of sense on its own.

You can just return a `new MatchReference` from that `getMatchReference()` method as that implements the interface.

src/Invoice.php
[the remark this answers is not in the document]
Wouter0100: That is exactly what I discovered. I think it wasn't documentation, but I think the Twinfield support confirmed that this would be the right solution.

src/Invoice.php
Wouter0100: Updated.

Wouter0100:
Feedback is processed :smile: :+1:.

willemstuursma:
Thanks @Wouter0100. Do you want us to release a new version with these changes, or wait for the other PR.

willemstuursma:
Thanks @Thijs-Riezebeek great feedback.

Wouter0100:
I'll open a PR soon. For now you should be able to do a release just fine. Thanks :smile:

PR #106, opened 2018-06-04
Added support for Customer Collect Mandate

RickJeroen:
Mandate which the debtor can collect with.

### DIFF
--- a/src/CustomerCollectMandate.php
+++ b/src/CustomerCollectMandate.php
@@ -60,10 +60,10 @@ class CustomerCollectMandate
     }
 
     /**
-     * @param \DateTimeInterface $date
+     * @param \DateTimeInterface|null $date
      * @return $this
      */
-    public function setSignatureDate(\DateTimeInterface $date): self
+    public function setSignatureDate(?\DateTimeInterface $date): self
     {
         $this->signatureDate = $date;
 
@@ -71,17 +71,17 @@ class CustomerCollectMandate
     }
 
     /**
-     * @param string $dateString
+     * @param string|null $dateString
      * @return $this
      * @throws Exception
      */
-    public function setSignatureDateFromString(string $dateString): self
+    public function setSignatureDateFromString(?string $dateString): self
     {
         if (!empty($dateString)) {
-            $this->setSignatureDate(Util::parseDate($dateString));
+            $dateString = Util::parseDate($dateString);
         }
 
-        return $this;
+        return $this->setSignatureDate($dateString);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


willemstuursma:
Thanks @RickJeroen! Would it be possible for you to add an integration test to https://github.com/php-twinfield/twinfield/blob/master/tests/IntegrationTests/CustomerIntegrationTest.php?

src/CustomerCollectMandate.php
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * @param string $dateString
+     * @return $this
+     * @throws Exception
+     */
+    public function setFirstRunDateFromString(string $dateString): self
+    {
+        if (!empty($dateString)) {
```
willemstuursma: I don't think the empty check is necessary if the argument is typehinted to `string`.

src/DomDocuments/CustomersDocument.php
```diff
@@ -103,6 +104,39 @@ public function addCustomer(Customer $customer): void
                 // Add the full element
                 $financialElement->appendChild($element);
             }
+
+            //check if collectmandate should be set
+            if ($customer->getCollectMandate() !== null) {
+
```
willemstuursma: I know it's not all best practice here in this class, but could you try to see if you can use some of the convenience methods in `\PhpTwinfield\DomDocuments\BaseDocument`?

src/Mappers/CustomerMapper.php
```diff
@@ -88,8 +88,34 @@ public static function map(Response $response)
...
+                $collectMandateTags = array(
+                    'id' => 'setID',
+                    'signaturedate' => 'setSignatureDateFromString',
+                    'firstrundate' => 'setFirstRunDateFromString',
+                );
+
+                $customer->setCollectMandate(new \PhpTwinfield\CustomerCollectMandate());
+
+                // Go through each collect mandate element and add to the assigned method
+                foreach ($collectMandateTags as $tag => $method) {
```
willemstuursma: Any of the convenience methods in `\PhpTwinfield\Mappers\BaseMapper` that can be useful here?

src/CustomerCollectMandate.php
RickJeroen: The Customer getter may return an empty string for signatureDate and firstRunDate.

willemstuursma: Then it should be nullable.

RickJeroen: True, just fixed it.

PR #111, opened 2018-07-10
Add support to list all offices when using AccessToken based authentication

manuelderuiter:
This pull request adds support to list all offices available to the current user authenticated via OAuth / OpenID. 

**Use case:**
We have a case where the customer wants to choose an office without giving a starting point.

**Solves:**
- 400 Bad Request when calling validateToken without an access token.
- Allow to instantiate OpenIdConnectAuthentication without an office present.

**TODO**
- Integration test: at the moment we don't have time to contribute an integration test to see what it does on all the other classes. We plan to do this later this week.

### DIFF
--- a/src/ApiConnectors/OfficeApiConnector.php
+++ b/src/ApiConnectors/OfficeApiConnector.php
@@ -4,6 +4,7 @@ namespace PhpTwinfield\ApiConnectors;
 
 use PhpTwinfield\Office;
 use PhpTwinfield\Services\FinderService;
+use PhpTwinfield\Request\Catalog\Office as OfficeRequestDocument;
 
 /**
  * A facade to make interaction with the the Twinfield service easier when trying to retrieve or send information about
@@ -16,6 +17,38 @@ use PhpTwinfield\Services\FinderService;
  */
 class OfficeApiConnector extends BaseApiConnector
 {
+    /**
+     * List the available offices when you are using the OAuth based authentication and don't have an office code yet.
+     * For more information following see.
+     *
+     * @see https://c3.twinfield.com/webservices/documentation/#/ApiReference/Types/XmlWebServices
+     * @throws \SoapFault
+     * @throws \PhpTwinfield\Exception
+     * @throws \ErrorException
+     * @return Office[] The offices found.
+     */
+    public function listAllWithoutOfficeCode(): array
+    {
+        $offices = [];
+        $document = new OfficeRequestDocument();
+        $response = $this->getProcessXmlService()->sendDocument($document);
+        $response->assertSuccessful();
+
+        /**
+         * @var \DOMElement $twinfieldOfficeResponse
+         */
+        foreach ($response->getResponseDocument()->firstChild->childNodes as $twinfieldOfficeResponse) {
+            $twinfieldOfficeContent = $twinfieldOfficeResponse->firstChild;
+
+            $office = new Office();
+            $office->setCode($twinfieldOfficeContent->textContent);
+            $office->setName($twinfieldOfficeResponse->attributes->getNamedItem("name")->textContent);
+            $offices[] = $office;
+        }
+
+        return $offices;
+    }
+
     /**
      * List all offices.
      *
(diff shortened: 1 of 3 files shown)


src/ApiConnectors/OfficeApiConnector.php
```diff
@@ -16,6 +17,38 @@
...
+    {
+        $offices = [];
+        $document = new OfficeRequestDocument();
+        $response = $this->getProcessXmlService()->sendDocument($document);
+        $response->assertSuccessful();
+
+        /**
+         * @var \DOMElement $twinfieldOfficeResponse
+         */
+        foreach ($response->getResponseDocument()->firstChild->childNodes as $twinfieldOfficeResponse) {
```
willemstuursma: I think this should be handled by a dedicated `Mapper` class.

src/Secure/OpenIdConnectAuthentication.php
```diff
@@ -95,8 +105,11 @@ protected function login(): void
     protected function validateToken(): array
     {
         $validationUrl    = "https://login.twinfield.com/auth/authentication/connect/accesstokenvalidation?token=";
+        $validationResult = false;
         try {
-            $validationResult = @file_get_contents($validationUrl . urlencode($this->accessToken));
+            if ($this->accessToken !== null) {
+                $validationResult = @file_get_contents($validationUrl . urlencode($this->accessToken));
```
willemstuursma: I don't think we need a try / catch here, but a check with `false`. Would you mind changing that? Maybe add the [http response header data](https://secure.php.net/manual/en/reserved.variables.httpresponseheader.php).

src/Secure/OpenIdConnectAuthentication.php
```diff
@@ -95,8 +105,11 @@ protected function login(): void
     protected function validateToken(): array
     {
         $validationUrl    = "https://login.twinfield.com/auth/authentication/connect/accesstokenvalidation?token=";
+        $validationResult = false;
```
willemstuursma: It seems that this can be simplified somewhat.

If there is no Office, the validationResult will always be an `OAuthException` right? Can you update the code here to reflect that?

willemstuursma:
@mvdpanne can you (briefly) look at these changes and give your opinion?

willemstuursma:
@manuelderuiter do you have time to improve this PR?

src/Secure/OpenIdConnectAuthentication.php
```diff
@@ -59,13 +62,20 @@ protected function getCluster(): ?string
 
     protected function getSoapHeaders()
```
mvdpanne: Perhaps a `OpenIdConnectAuthentication::setOffice` method could be added, such that you can always set the Office afterwards. This method would then have to set `AuthenticatedConnection::$authenticatedClients` to an empty array, to make sure new clients (with the `CompanyCode` header) will be created afterwards.

src/Secure/OpenIdConnectAuthentication.php
mvdpanne: I'm pretty sure the `Office` has no relation to the process of fetching the access token (OAuth2 flow) nor validating it (the url only takes the access token as argument, and no headers).

So, you can have a valid token without supplying an Office instance.

src/Secure/OpenIdConnectAuthentication.php
mvdpanne: I guess so. I cannot remember why I added the catch there. I did add a `@`, so it should not be able to throw exceptions.

Oh, and is that check: `if ($this->accessToken !== null)` really necessary?

manuelderuiter:
@willemstuursma I've been quite busy these last days. Tomorrow I will make some time available to process these changes.

src/Secure/OpenIdConnectAuthentication.php
manuelderuiter: When I activated Charles to see what we were sending to Twinfield I noticed that we got a HTTP 400 Bad Request reply and that $accessToken was empty at all first calls. When we initiate the `OpenIdConnectAuthentication` class we provide a refresh token.

Call stack from what I remember: login -> **validateToken** -> refreshToken -> validateToken.

willemstuursma:
Cheers @manuelderuiter once it's all polished up I will merge and release a new version.

src/Secure/OpenIdConnectAuthentication.php
mvdpanne: Ah, I see. I think at the time I reasoned that the failed validation request was 'fine', as it then refreshes the token (as indicated by your call stack).

Perhaps it would be neater to add the following check to the `login` function instead, before the try-catch:
```
if ($this->accessToken === null) {
    $this->refreshToken();
}
```

PR #112, opened 2018-07-16
Fix typo's in getFreetext()

japaveh:
There seems to be a small copy-paste error in the $object->getFreetextX() which results that the getFreetext() don't match with the statement below.


willemstuursma:
Cheers!

PR #114, opened 2018-07-17
Add support for Cash Transactions

Versleijen:
This pull request adds support for Cash Transactions in Twinfield.

### DIFF
--- a/src/CashTransaction.php
+++ b/src/CashTransaction.php
@@ -6,33 +6,19 @@ use Money\Currency;
 use Money\Money;
 use PhpTwinfield\Enums\DebitCredit;
 use PhpTwinfield\Enums\LineType;
+use PhpTwinfield\Transactions\TransactionFields\StartAndCloseValueFields;
 use PhpTwinfield\Transactions\TransactionFields\StatementNumberField;
 use PhpTwinfield\Transactions\TransactionLine;
-use Webmozart\Assert\Assert;
 
 /**
- * NOTE:    Can not use the trait "PhpTwinfield\Transactions\TransactionFields\StartAndCloseValueFields" because the
- *          method "getCurrency" is not compatible with the method "getCurrency" of ""PhpTwinfield\BaseTransaction"".
- *
  * @link https://c3.twinfield.com/webservices/documentation/#/ApiReference/Transactions/CashTransactions
  */
 class CashTransaction extends BaseTransaction
 {
     use StatementNumberField;
-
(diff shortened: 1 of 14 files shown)


src/CashTransaction.php
```diff
@@ -0,0 +1,127 @@
+<?php
```
willemstuursma: is this (partially) copied from another class? Please see if you can move some code into traits then.

src/CashTransactionLine.php
```diff
@@ -0,0 +1,264 @@
+<?php
```
willemstuursma: Same for this class: is this (partially) copied from another class? Please see if you can move some code into traits then.

src/CashTransaction.php
Versleijen: Yes it is. I will look into it!

src/CashTransactionLine.php
Versleijen: Yes it is. I will look into it!

willemstuursma:
@Versleijen do you have time to make the changes to this PR?

src/CashTransaction.php
Versleijen: @willemstuursma The reason why I copied some code is because the methods *getCurrency* and *setCurrency* of the trait *StartAndCloseValueFields* are not compatible with the methods of *PhpTwinfield\BaseTransaction*. If I change the methods in *BaseTransaction* I'm afraid I will break existing implementations. Any idea how to solve this?

src/CashTransactionLine.php
Versleijen: @willemstuursma Here I followed the documentation of Twinfield. As far I can see, according to the documentation, each of the transactions lines (Bank, Cash, Sales) are slightly different in which value is valid/allowed for each line type. Do you have an example where we could share code?

src/CashTransaction.php
willemstuursma: Oh that sucks! It doesn't use any Currency value objects. 

To me, updating `PhpTwinfield\BaseTransaction` is fine. We'll release a new minor version with a changelog instead of a bugfix version.

src/CashTransactionLine.php
willemstuursma: It's too bad they are all slightly different. Thanks for looking into it.

willemstuursma:
Thanks. Weird that `phpstan` crashed.

PR #115, opened 2018-07-17
removeAddress() method in the Customer class.

mikedutchtelco:
I would like to use the removeAddress() method in the Customer class but when I try to use it, it give me an error. After searching in the method I noticed that the 'addresses' property in this method was misspelled two times (two different ways).

I've changed the name of the properties and it now works fine for me.

### DIFF
--- a/src/Customer.php
+++ b/src/Customer.php
@@ -338,6 +338,16 @@ class Customer
         return $this;
     }
 
+    public function removeAddress($index)
+    {
+        if (array_key_exists($index, $this->addresses)) {
+            unset($this->addresses[$index]);
+            return true;
+        } else {
+            return false;
+        }
+    }
+
     public function getBanks()
     {
         return $this->banks;


willemstuursma:
Why do you need this method? Why not create a new Customer with the correct address?

mikedutchtelco:
In my opinion it is strange to create a whole new Customer if you just want to update or delete an Address.
And this method did always exists, why is it removed in the latest version?

willemstuursma:
> This method did always exists, why is it removed in the latest version?

It didn't work at all. 

> In my opinion it is strange to create a whole new Customer if you just want to update or delete an Address.

This is like, retrieving a Customer from Twinfield, and then removing an address, and then sending the Customer to Twinfield again right?

I meant on an object but if the above is the usecase, it looks valid.

src/Customer.php
```diff
@@ -338,6 +338,16 @@ public function addAddress(CustomerAddress $address)
         return $this;
     }
 
+    public function removeAddress($index)
+    {
+        if (array_key_exists($index, $this->addresses)) {
+            unset($this->addresses[$index]);
+            return true;
+        } else {
```
willemstuursma: No `else` is needed here, because of the return. 

Can you add type hints, e.g. int for the parameter index and bool for the method return value.

PR #116, opened 2018-08-14
Twinfield expects string representation of boolean ("true"/"false")

Versleijen:
Twinfield expects string representation of boolean ("true"/"false") instead of 1 or 0 for "default" attribute in XML for customer address and customer bank element.

### DIFF
--- a/src/CustomerAddress.php
+++ b/src/CustomerAddress.php
@@ -53,12 +53,19 @@ class CustomerAddress
         return $this;
     }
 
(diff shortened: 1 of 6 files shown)


willemstuursma:
Do we have some nice constants for these `true` / `false` strings?

Can you add a type hint to `getDefault()` and `setDefault()`?

willemstuursma:
Cheers, this is great!

Issue #122, opened 2018-11-06
OpenID Oauth2.0 Authentication


gerkevgarderen:
The scopes `'twf.user'` is necessary for some requests. But adding the scope `twf.user` will result in the error `invalid_scope` - earlier the assumption was it is not in use by Twinfield.

However, if `openid` is added to the list of scopes in `getAuthorizationUrl()`, no error will be raised and the application works perfectly.

In order to get this to work, I made some changes to the code in Secure/Provider/OAuthProvider.php, around the old line 107-109 to remove the throw Exception, and allow twf.user as scope, so long as openid is added as scope as well.

With these added, the AccessToken works completely as a replacement for sessions.

willemstuursma:
Sounds great! Can you open a PR with your changes?

PR #123, opened 2018-11-06
Added support for twf.user scope through OpenID scope

gerkevgarderen:
Updated the OAuthProvider login to allow the twf.user scope by adding OpenID scope if it is not yet included.

### DIFF
--- a/src/Secure/Provider/OAuthProvider.php
+++ b/src/Secure/Provider/OAuthProvider.php
@@ -17,8 +17,12 @@ class OAuthProvider extends AbstractProvider
     use BearerAuthorizationTrait;
 
     /**
-     * Please note: This scope only works when the SCOPE_OPEN_ID scope is also present.
-     *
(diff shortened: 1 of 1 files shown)


gerkevgarderen:
Based on #122

src/Secure/Provider/OAuthProvider.php
```diff
@@ -44,18 +48,6 @@ class OAuthProvider extends AbstractProvider
      */
     public const SCOPE_ORGANISATION_USER = 'twf.organisationUser';
 
-    /**
-     * This scope is required in order to receive and use refresh tokens.
-     */
-    public const SCOPE_OFFLINE_ACCESS = "offline_access";
```
willemstuursma: Why did you remove this constant? Is something wrong with it?

gerkevgarderen:
Turns out to be changes in an older version of the file. Checking again.

PR #126, opened 2019-02-18
Add offline_access and organisation scopes

### LINKED ISSUE #125: OAuthProvider doesn't use offline_access and organisation scopes
I've noticed that the `OAuthProvider` doesn't use the `offline_access` scope by default. This results in `$accessToken->getRefreshToken()` returning `NULL`. It looks like the provider only uses these scopes by default:

- twf.organisationUser
- openid

The [Twinfield documentation](https://c3.twinfield.com/webservices/documentation/#/ApiReference/Authentication/OpenIdConnect) states that the `offline_access` scope has to be present in order to receive a refresh token. After adding `self::SCOPE_OFFLINE_ACCESS` to the array that `$provider->getDefaultScopes()` returns, a refresh token is available.

EDIT:
After adding `self::SCOPE_OFFLINE_ACCESS` I got an undefined index notice that `twf.clusterUrl` wasn't set. This variable belongs to the `self::SCOPE_ORGANISATION` scope and  is used in the `login()` method of the `OpenIdConnectAuthentication` class.

svscherpenseel:
Added the offline_access and organisation scopes to the array of default scopes in order to receive a refresh token and the cluster URL respectively, according to the [Twinfield documentation](https://c3.twinfield.com/webservices/documentation/#/ApiReference/Authentication/OpenIdConnect).

Closes #125

### DIFF
--- a/src/Secure/Provider/OAuthProvider.php
+++ b/src/Secure/Provider/OAuthProvider.php
@@ -88,12 +88,18 @@ class OAuthProvider extends AbstractProvider
     /**
      * Returns the default scopes used by this provider.
      *
-     * This should only be the scopes that are required to request the details of the resource owner, rather than all
-     * the available scopes.
+     * Besides the ORGANISATION_USER and OPEN_ID scopes, the OFFLINE_ACCESS and ORGANISATION scopes have to
+     * be used in order to receive a refresh token and cluster URL. These are necessary to refresh the access
+     * token and make requests to the correct endpoint.
      */
     protected function getDefaultScopes(): array
     {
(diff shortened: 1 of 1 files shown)


willemstuursma:
Can you add a link to the issue and if possible Twinfield docs?

svscherpenseel:
@willemstuursma done!

willemstuursma:
Did you push your changes @svscherpenseel? Shows only 1 commit from yesterday.

svscherpenseel:
That's correct @willemstuursma. That's the only thing it takes to receive a refresh token and the cluster URL.

willemstuursma:
Ah now I see. I meant in the code, sorry, was not very clear. That's why I expected another commit.

PR #141, opened 2019-04-28
Remove BankTransactionApi and move functionality to TransactionsApi

iranl:
Considering BankTransactions are identical to CashTransactions they should be treated the same by the library. This constitutes removing the BankTransactionsApiConnector en adding all functionality in exactly the same way as CashTransactions to the TransactionsApiConnector.

Amounts to a breaking change for those using the BankTransactionsApiConnector.


iranl:
Superseded by #142

PR #142, opened 2019-05-09
[WIP] Rewrite/Extend all resources

iranl:
Supersedes all earlier PR's by me. Implements all requested changes in those PR's.
Also fixes some other open issues.

Unfortunately, mainly because everything is moved to traits, it is now impossible to separate the changes into different PR's for each resource.

Changes (non exhaustive list, applies to alle resources)
- All fields (including fields that are only used once) moved to traits in the Fields namespace. Re-use traits whenever possible (same exact element name and type in Twinfield API).
- Created Enums for every field with an exhaustive list
- Make a PhpTwinfield object of elements whenever possible (in the same way the office code field is turned into an Office object)
- Implement all DOM attributes from element fields (name, shortname, dimensiontype, locked, inherit etc,)
- Rewrote all ApiConnectors, Mappers and Documents to mimic Transactions
- Added resources (see below)

| Component                                                                                                                         | get()              | listAll()          | send()             | delete()           |  Mapper            |
| ----------------------------------------------------------------------------------------------------------------------------------| :----------------: | :----------------: | :----------------: | :----------------: | :----------------: |
| [Activities](https://c3.twinfield.com/webservices/documentation/#/ApiReference/Masters/Activities)                                | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| [Articles](https://c3.twinfield.com/webservices/documentation/#/ApiReference/Masters/Articles)                                    | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| [Asset Methods](https://c3.twinfield.com/webservices/documentation/#/ApiReference/Masters/AssetMethods)                           | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| [Browse Data](https://c3.twinfield.com/webservices/documentation/#/ApiReference/Request/BrowseData)                               | :white_check_mark: |                    |                    |                    | :white_check_mark: |
| Cash and Bank Books                                                                                                               |                    | :white_check_mark: |                    |                    | :white_check_mark: |
| [Cost Centers](https://c3.twinfield.com/webservices/documentation/#/ApiReference/Masters/CostCenters)                             | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| Countries                                                                                                                         |                    | :white_check_mark: |                    |                    | :white_check_mark: |
[...]

PR #144, opened 2019-06-10
Base modifications

iranl:
Comprises part of the changes in #142

- Add additions to current base files (e.g. BaseApiConnector, BaseDocuments, BaseMapper)
- Other small modifications that have no effect on existing functionality

BaseApiConnector
- Adds getConnection function which is used to call other ApiConnectors from within an ApiConnector
- Adds convertOptionsToArrayOfString function to convert a simple $options array to a Twinfield acceptable ArrayOfString when using the Finder
- Adds mapListAll function to allow easy mapping of listAll results to objects

BaseDocument
- Extends CreateNodeWithTextContent function to optionally add attributes to a node

BaseMapper
- Extends getField function to optionally allow checking for (error) messages on the field
- Adds getAttribute function to read attributes of a node
- Adds parse functions for Boolean, Date, DateTime, Enum Money and Object fields 

Read
- Adds some child functionality to Read, so these common functions can later be removed from the children

OpenIdConnectAuthentication
- Extends __construct function with the optional ability to connect using a known access token and access cluster for quicker connections

BaseService
- Modification to switch WSDL from __call to __soapcall because of deprecation of the former (see comments in #142)

Other
- Removal of unnecessary whitespaces and typos


src/ApiConnectors/BaseApiConnector.php
```diff
@@ -47,6 +47,11 @@ public function __construct(AuthenticatedConnection $connection)
         $this->connection = $connection;
     }
 
+    public function getConnection()
```
willemstuursma: Could you add a typehint for the return type? Why is this method needed?

src/ApiConnectors/BaseApiConnector.php
```diff
@@ -161,4 +166,70 @@ protected function getFinderService(): FinderService
...
+     *
+     * @param array $options
+     * @param array|null $forcedOptions
+     * @return array
+     * @throws Exception
+     */
+    public function convertOptionsToArrayOfString(array $options, array $forcedOptions = null): array {
+        if (isset($options['ArrayOfString'])) {
+            return $options;
+        } else {
```
willemstuursma: These is no need to have an `else` if the if body has a `return` statement, can you change this to use less indenting?

src/ApiConnectors/BaseApiConnector.php
```diff
@@ -161,4 +166,70 @@ protected function getFinderService(): FinderService
...
+            $optionsArrayOfString = array('ArrayOfString' => array());
```
willemstuursma: Can you use the modern `[]` syntax for creating arrays?

src/ApiConnectors/BaseApiConnector.php
```diff
@@ -161,4 +166,70 @@ protected function getFinderService(): FinderService
...
+
+    /**
+     * Convert options array to an ArrayOfString which is accepted by Twinfield.
+     *
+     * @param array $options
+     * @param array|null $forcedOptions
+     * @return array
+     * @throws Exception
+     */
+    public function convertOptionsToArrayOfString(array $options, array $forcedOptions = null): array {
```
willemstuursma: `$forcedOptions` can be defaulted to an `[]`, that allows you to skip the isset() check too.

src/ApiConnectors/BaseApiConnector.php
```diff
@@ -161,4 +166,70 @@ protected function getFinderService(): FinderService
     {
         return $this->connection->getAuthenticatedClient(Services::FINDER());
     }
-}
+
+    /**
+     * Convert options array to an ArrayOfString which is accepted by Twinfield.
+     *
+     * @param array $options
+     * @param array|null $forcedOptions
```
willemstuursma: Can you add a comment to explain what the 2nd argument does?

src/DomDocuments/BaseDocument.php
```diff
@@ -156,13 +156,24 @@ protected function appendPerformanceTypeFields(\DOMElement $element, $object): v
...
      *
      * @param string $tag
-     * @param string $textContent
+     * @param string|null $textContent
+     * @param $object
+     * @param array|null $attributes
      * @return \DOMElement
      */
-    final protected function createNodeWithTextContent(string $tag, string $textContent): \DOMElement
+    final protected function createNodeWithTextContent(string $tag, ?string $textContent, $object = null, array $attributes = null): \DOMElement
```
willemstuursma: Again, the last argument can default to `[]`.

src/DomDocuments/BaseDocument.php
```diff
@@ -156,13 +156,24 @@ protected function appendPerformanceTypeFields(\DOMElement $element, $object): v
      * Use this instead of createElement().
      *
      * @param string $tag
-     * @param string $textContent
+     * @param string|null $textContent
+     * @param $object
+     * @param array|null $attributes
```
willemstuursma: Can you add how this argument would be used? Maybe rename to `$methodToAttributeMap`?

src/Mappers/BaseMapper.php
```diff
@@ -62,17 +64,158 @@ protected static function getValueFromTag(\DOMDocument $document, string $tag):
...
+        }
+
         if ($fieldElement->textContent === "") {
             return null;
         }
 
         return $fieldElement->textContent;
     }
+    
+    protected static function checkForMessage($object, \DOMElement $element): void
```
willemstuursma: The object should have a typehint or implement an interface, e.g. `HasMessage`.

src/Mappers/BaseMapper.php
```diff
@@ -62,17 +64,158 @@ protected static function getValueFromTag(\DOMDocument $document, string $tag):
...
+        }
+
+        if ($fieldElement->getAttribute($attributeName) === "") {
+            return null;
+        }
+
+        return $fieldElement->getAttribute($attributeName);
+    }
+
+    protected static function parseBooleanAttribute(?string $value): ?bool
```
willemstuursma: Nice, but there is a method in `Util` that does this already?

src/Mappers/BaseMapper.php
```diff
@@ -62,17 +64,158 @@ protected static function getValueFromTag(\DOMDocument $document, string $tag):
...
+    {
+        return filter_var($value, FILTER_VALIDATE_BOOLEAN);
+    }
+
+    protected static function parseDateAttribute(?string $value): ?\DateTimeImmutable
+    {
+        if ((bool)strtotime($value)) {
```
willemstuursma: I think a `false !== strtotime()` would be better.

src/Mappers/BaseMapper.php
```diff
@@ -62,17 +64,158 @@ protected static function getValueFromTag(\DOMDocument $document, string $tag):
...
+            return Util::parseDate($value);
+        } else {
+            return null;
```
willemstuursma: Please remove the unneeded level of indentation.

src/Mappers/BaseMapper.php
```diff
@@ -62,17 +64,158 @@ protected static function getValueFromTag(\DOMDocument $document, string $tag):
...
+    protected static function parseDateTimeAttribute(?string $value): ?\DateTimeImmutable
+    {
+        if ((bool)strtotime($value)) {
+            return Util::parseDateTime($value);
+        } else {
+            return null;
+        }
+    }
+
+    protected static function parseEnumAttribute(string $enumName, ?string $value)
```
willemstuursma: I think it would be better to pass in the entire class name, that way you can use `::class` and can easily find usages.

These string classname concatination patterns make that very hard and offer little value.

src/Mappers/BaseMapper.php
```diff
@@ -62,17 +64,158 @@ protected static function getValueFromTag(\DOMDocument $document, string $tag):
...
+            $classReflex = new \ReflectionClass($enum);
+            $classConstants = $classReflex->getConstants();
+
+            foreach ($classConstants as $classConstant) {
+                if ($value == $classConstant) {
+                    return new $enum($value);
+                }
+            }
+        } catch (\ReflectionException $e) {
+            return null;
```
willemstuursma: If the class does not exist, it should be an error. Its probably a typo then, so the developer should be alerted.

src/Mappers/BaseMapper.php
```diff
@@ -62,17 +64,158 @@ protected static function getValueFromTag(\DOMDocument $document, string $tag):
...
+        }
+
+        $enum = "\\PhpTwinfield\\Enums\\" . $enumName;
+
+        try {
+            $classReflex = new \ReflectionClass($enum);
+            $classConstants = $classReflex->getConstants();
+
+            foreach ($classConstants as $classConstant) {
+                if ($value == $classConstant) {
```
willemstuursma: Should use strict equality, Enum, supports 0, false and null as valid values.

willemstuursma: Alternatively, you can just try to construct them and if it fails, return null (or fail).

src/Mappers/BaseMapper.php
```diff
@@ -62,17 +64,158 @@ protected static function getValueFromTag(\DOMDocument $document, string $tag):
...
+        return null;
+    }
+
+    protected static function parseMoneyAttribute(?float $value): ?Money
+    {
+        if ($value === null) {
+            return null;
+        }
+
+        return Util::parseMoney($value, new Currency('EUR'));
```
willemstuursma: This is not good, we use Twinfield in a many currencies environment and a special treatment of EUR leads to bugs.

Please allow to currency to be passed as an argument, or remove this method.

src/Mappers/BaseMapper.php
```diff
@@ -62,17 +64,158 @@ protected static function getValueFromTag(\DOMDocument $document, string $tag):
...
+    }
+
+    protected static function parseObjectAttribute(string $className, $object, \DOMElement $element, string $fieldTagName, array $attributes = null)
```
willemstuursma: Based on the earlier comments, can you update this method?

src/Mappers/BaseMapper.php
```diff
@@ -62,17 +64,158 @@ protected static function getValueFromTag(\DOMDocument $document, string $tag):
...
+                    $className = "CostCenter";
+                    break;
+                case "PNL":
+                    $className = "GeneralLedger";
+                    break;
+                case "PRJ":
+                    $className = "Project";
+                    break;
+                default:
+                    return null;
```
willemstuursma: This should be an `\InvalidArgumentException`

src/Mappers/BaseMapper.php
```diff
@@ -62,17 +64,158 @@ protected static function getValueFromTag(\DOMDocument $document, string $tag):
...
+            }
+        }
+
+        $class = "\\PhpTwinfield\\" . $className;
+
+        $object2 = new $class();
```
willemstuursma: Can you add a typehint: `/** @var SomeClassWithMethodsetCode $object2 */`

src/Request/Read/BrowseDefinition.php
```diff
@@ -2,17 +2,15 @@
...
 class BrowseDefinition extends Read
 {
     /**
      * Sets office and code if they are present.
      *
      * @param string $code
-     * @param Office|null $office
+     * @param $office
      */
-    public function __construct(string $code, Office $office = null)
```
willemstuursma: Please don't remove the typehint?

src/Request/Read/Transaction.php
```diff
@@ -1,8 +1,6 @@
 <?php
 namespace PhpTwinfield\Request\Read;
 
-use PhpTwinfield\Office;
```
willemstuursma: Can you put all the Office typehints back?

willemstuursma (changes requested):
Thank you so much @iranl for taking the time to create this PR.

Some generic feedback:

* `[]` can be the default argument for array typed parameters, this allows you to remove the `isset` check
* The usage of some method parameters is not clear to me, can you update the relevant phpdocs.
* Please do not treat EUR currency special, always have it be made explicit
* Full class names are superior to shortened classnames: they can be autocompleted by the IDE, type checked and their usages can easily be found
* The `Office` type has been removed in some places, I think this is a merge error?

In my company we use PhpStorm and we rely heavily on type hints to prevent bugs, hence the focus on types. Basically, any method call that you cannot click in the IDE is frowned upon. 

Please let me know if you have any questions or would like to discuss.

src/ApiConnectors/BaseApiConnector.php
iranl: The modified` InvoicesDocument addInvoice` function uses the `InvoiceTypeApiConnector` and `ArticleApiConnector` to request additional information on the selected `InvoiceType` and `Articles` in the `Invoice`, which means it needs access to the current connection. The `InvoiceApiConnector sendAll` function calls the parent `BaseApiConnector getConnection` function and passes the connection as an argument to `addInvoice`.

Typehint added

src/ApiConnectors/BaseApiConnector.php
iranl: Done

src/ApiConnectors/BaseApiConnector.php
iranl: Done

src/ApiConnectors/BaseApiConnector.php
iranl: Done

src/ApiConnectors/BaseApiConnector.php
iranl: In some cases you are not allowed to change certain options (such as the dimtype, which should always be DEB when using `CustomerApiConnector ListAll`), in which case the `$forcedOptions` parameter will be set by the `ApiConnector` for this option, which will override any user settings in `$options`

Added to description

src/DomDocuments/BaseDocument.php
iranl: Done

src/DomDocuments/BaseDocument.php
iranl: Argument is used as follows:

`$this->createNodeWithTextContent('billable', $quantity->getBillableToString(), $quantity, array('locked' => 'getBillableLockedToString'))`

Method renamed

src/Mappers/BaseMapper.php
iranl: True, but I like this one better as it uses built in PHP functionality and is more versatile. Could probably be moved to `Util`, but functionality is not 1 on 1 comparable and could then have the potential of breaking something.

src/Mappers/BaseMapper.php
iranl: Done

src/Mappers/BaseMapper.php
iranl: Done

src/Mappers/BaseMapper.php
iranl: Done

src/Mappers/BaseMapper.php
iranl: Done, now throws exception

src/Mappers/BaseMapper.php
iranl: This is no specials treatment op EUR. There are no value fields in Twinfield directly linked to a currency, so whichever currency is used here has zero effect on functionality.

Money for PHP just needs some currency to construct an object. Could be any ISO currency.

src/Mappers/BaseMapper.php
iranl: Done

src/Mappers/BaseMapper.php
iranl: Done

src/Mappers/BaseMapper.php
iranl: Done

src/Request/Read/BrowseDefinition.php
iranl: Moving `setOffice` to `Read` gave errors in either `setOffice` in `Browse data`/`Transactions` or in most other objects because of correct type hinting in `Browse Data` and `Transactions` and bad type hinting in the other objects. Eventually chose the shortcut of removing typehints. Now properly fixed by fixing the bad type hinting

src/Request/Read/Transaction.php
iranl: See above

src/Mappers/BaseMapper.php
iranl: Changed to:

`try {
            return new $enumClass($value);
        } catch (\Exception $e) {
             return null;
        }`

src/Mappers/BaseMapper.php
iranl: I'm not really up to date on using interfaces and typehinting for multiple classes. Can you give me some help on this? Or is this something that we can address at a later stage?

src/ApiConnectors/BaseApiConnector.php
```diff
@@ -169,53 +175,52 @@ protected function getFinderService(): FinderService
...
+        return $optionsArrayOfString;
     }
 
     /**
      * Map the response of a listAll to an array of the requested class
      *
-     * @param string $className
+     * @param string $objectClass
      * @param $data
      * @param array $objectListAllTags
```
willemstuursma: This is still missing an explanation of its usage.

src/Mappers/BaseMapper.php
willemstuursma: I'm not sure if I understand this correctly - Twinfield does not pass a currency for some amounts?

src/Mappers/BaseMapper.php
```diff
@@ -166,54 +164,51 @@ protected static function parseMoneyAttribute(?float $value): ?Money
         return Util::parseMoney($value, new Currency('EUR'));
     }
 
-    protected static function parseObjectAttribute(string $className, $object, \DOMElement $element, string $fieldTagName, array $attributes = null)
+    /** @var SomeClassWithMethodsetCode $object2 */
+    protected static function parseObjectAttribute(string $objectClass, $object, \DOMElement $element, string $fieldTagName, array $attributes = [])
     {
-        if ($className == "DimensionGroupDimension" || $className == "UnknownDimension") {
-            if ($className == "DimensionGroupDimension") {
+        if ($objectClass == "DimensionGroupDimension" || $objectClass == "UnknownDimension") {
```
willemstuursma: Should these strings be updated to class references (`::class`)?

src/Mappers/BaseMapper.php
willemstuursma: Well, this method calls `addMessage()` on the object, so it cannot take _any_ object right. It has to take an object that supports messages and the required method call. 

So ideally, these objects would have an interface `HasMessage` that defines the method `addMessage(Message $message): void` so that you can use this typehint in the method parameter list. 

You can read a bit about interfaces here: https://www.zentut.com/php-tutorial/php-interface/ Please ignore the crappy naming in the article, interfaces should have names like `SupportFoo`, `CanBar` or even `IsBazable` to indicate the class supports some _functionality_.

src/Mappers/BaseMapper.php
willemstuursma: I'm not entirely convinced - I think that it for new functionality it would be hard to decide which boolean parsing function should be used. The existence of two is confusing if there is no clear guideline to pick one over the other.

Please consolidate into one method, its fine to change the one in Util, that's why we have tests.

willemstuursma (commented):
Thanks @iranl! We're making good progress.

src/Mappers/BaseMapper.php
iranl: Twinfield does not pass currency for **any** amount

Money for PHP is used when a field type in the API Documentation equals money. For example in Articles (from the API documentation):

Name | Type | Description
-- | -- | --
unitspriceexcl | money | Price excluding VAT

If you look at the XML the Money Type equals just a decimal number in Twinfield:

`<unitspriceexcl>750.00</unitspriceexcl>`

It does not include or involve a Currency.

The most basic usage in the library would be to just cast these decimals als float instead of using Money for PHP, but this would mean you don't get access to the added functionality (such as add, subtract, formater etc.). Money for PHP needs some currency code to construct so any currency code can be used, I used EUR because it was already used this way in `Transactions.` Currency operations within the money object should not be performed.

src/Mappers/BaseMapper.php
iranl: No, they cannot, as they are non existent stubs that tell this function to determine the real class from an argument contained in the XML element or sibling XML element.

`parseObjectAttribute` is usually used when the Twinfield entity contained in the XML is static (such as the <user> element which will always contain a user (or nothing)). In this case `::class` is used when calling `parseObjectAttribute`

In DimensionGroupDimension, CustomerFinancials, FixedAssetTransactionLine and SupplierFinancials there are fields that can contain multiple or any of the 7 dimensionTypes. In these cases` $className == "DimensionGroupDimension"` or `$className == "UnknownDimension"` is used to tell the function to change `$objectClass` to the real class name in the following `switch`

src/ApiConnectors/BaseApiConnector.php
iranl: Done

src/Mappers/BaseMapper.php
iranl: Moved new method to Util

src/Mappers/BaseMapper.php
iranl: Done and believe it works as intended now

iranl:
@willemstuursma Moving along quite nicely. Thanks for all the reviewing!

src/Mappers/BaseMapper.php
willemstuursma: Isn't it in the header part of the XML document? https://c3.twinfield.com/webservices/documentation/#/ApiReference/SalesInvoices

iranl: A currency is sometimes part of the header or returned elsewhere in an entity (but more often not), but is not directly linked to individual amounts (in an attribute field on the field for example).

iranl: Another thing to realise is that Twinfield Currency codes don't have to be ISO Currency codes, they can be anything you like. If you try to set a non existing ISO currency in Money you will get an error.

I'm also ok with just removing Money for PHP entirely and treating money fields in Twinfield as floats. Is probably the cleanest solution.

iranl: @willemstuursma can you comment on this? My preference would be to remove Money for PHP entirely. 

After deciding on this, I think we might be ready to continue with the next bit of #142.

willemstuursma: Hi @iranl. We want to keep MoneyPHP\Money for sure. 

I think you should delete this method and rely on the method in Util, and always pass a currency. But it depends on the situation where it comes from.

iranl: Did some more testing and MoneyPHP\Money does in fact accept non ISO Currencies as input. So that's one (non-existent) problem out of the way.

This function in BaseMapper does have a use though as the function in Util does not accept null input for value or currency (which are possibilities when reading from Twinfield). This functions filters those out (and returns null) and passes all other input to the function in Util.

So I believe the function should stay and I have modified it to take a currency string as input instead of hardcoding it to EUR. 

There are some problems along the way with currencies though (such as values in the reporting currency instead of the base currency). But those have solutions in other places.

iranl: Added a function to BaseMapper to retrieve the Base and Reporting currency from the selected office for use in the specific mappers when working with currencies that are not linked to the currency in the specific object or when there is no currency field in the object

src/Mappers/BaseMapper.php
iranl: Moved this part of the parseObjectAttribute function to a different function, seems like a cleaner solution

iranl:
@willemstuursma Do you have any unsolved requests concerning the current PR?

If not, I would like to continue with the next part of #142

willemstuursma:
Sorry @iranl, I noticed some changes but it is really super busy in my day job, so I didn't have time to do any further reviewing yet.

iranl:
@willemstuursma no problem. Just thought I'd check you weren't waiting on more changes from me. Will continue the conversation when you have the time.

ericclaeren:
Hi, 

I'm new to this package, running into the access token issue where it's logging in for each request while I already have a valid token, which results in quite some additional requests.

I could help you with the access token part, although it would be breaking for current implementations, we could use AccessTokenInterface to pass the access + refresh token + expires to let this package check if it's valid and requesting a new access token when it's invalid. Or use a setter for it,

Or possible add a tokenRefresh callback method to call when token is refreshed so it can be stored externally. Because there's no way to get the access token from the current connection.

Don't mean to barge in, but would love have this feature :) and can get some time from my company to help with this.

Cheers, Eric

PR #145, opened 2019-06-14
Optional sort fields

John5:
When using browse code 690 Twinfield will return an error when a sort fields are specified in the request. Without sort fields it works. The PHP Twinfield library currently requires sort fields. This change makes the sort fields optional.

### DIFF
--- a/src/BrowseDataRow.php
+++ b/src/BrowseDataRow.php
@@ -57,7 +57,7 @@ class BrowseDataRow
      * @param string $code
      * @return BrowseDataRow
      */
-    public function setCode(string $code): BrowseDataRow
+    public function setCode(?string $code): BrowseDataRow
     {
         $this->code = $code;
         return $this;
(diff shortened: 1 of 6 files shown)


src/BrowseDataRow.php
```diff
@@ -57,7 +57,7 @@ public function getCode(): string
      * @param string $code
```
iranl: `@param string $code` should be changed to `@param string|null $code`

src/BrowseDataRow.php
```diff
@@ -75,7 +75,7 @@ public function getNumber(): int
      * @param int $number
```
iranl: `@param int $number` should be changed to `@param int|null $number`

src/BrowseDataRow.php
```diff
@@ -93,7 +93,7 @@ public function getLine(): int
      * @param int $line
```
iranl: `@param int $line` should be changed to `@param int|null $line`

PR #147, opened 2019-06-14
Optional date time patch

John5:
When Twinfield returns an empty date it should be converted to NULL.


src/ApiConnectors/BrowseDataApiConnector.php
```diff
@@ -69,7 +69,9 @@ public function getBrowseData(string $code, array $columns, array $sortFields =
     {
         Assert::minCount($columns, 1);
         Assert::allIsInstanceOf($columns, BrowseColumn::class);
-        Assert::allIsInstanceOf($sortFields, BrowseSortField::class);
+        if(count($sortFields)) {
```
willemstuursma: This isn't really needed right - an empty collection of $sortFields are also all of the `BrowseSortField` class.

PR #120, opened 2018-10-02
Remove duedate from SalesTransaction.

jelleroorda:
Removes the duedate from sales transactions, since the Twinfield API doesn't allow the duedate in the header anymore (`Vervaldatum is niet toegestaan in de koptekst.`).

Code to reproduce this issue

```php
$transaction = new SalesTransaction;
$transaction->setDestiny(Destiny::TEMPORARY())
    ->setCurrency($invoice->getCurrency())
    ->setDateFromString($invoice->getInvoiceDate()->format("Ymd"))
    ->setPeriod($invoice->getInvoiceDate()->format("Y/m"))
    ->setDueDateFromString($invoice->getDueDate()->format("Ymd"))
```

Will yield the following error:
```
Argument 1 passed to PhpTwinfield\SalesTransaction::setDueDateFromString() must be of the type string, null given, called in project/vendor/php-twinfield/twinfield/src/Mappers/TransactionMapper.php on line 107
```
Removing the duedate from `SalesTransaction` will fix this issue. Code not removed from the `TransactionMapper` since it might be used by other Transactions.

PR #149, opened 2019-07-08
Remove due date field from SalesTransaction

wmouwen:
Replaces stale PR #120


willemstuursma:
Thanks!

niektenhoopen:
I ran into the same issue today and am surprised that it is fixed in dev-master but not in the latest version 2.7.0. Any ideas when it will be released in (any?) version @willemstuursma ? I am uncomfortable using dev-master in production projects.

PR #153, opened 2019-08-16
Reset login when resetting webservice connection

fschalkwijk:
When using the webservice authenticator in a queue runner I often receive the error "Your logon credentials are not valid anymore. Try to log on again.". There is a catch in BaseApiConnector::sendXmlDocument which retries the request with a fresh connection, but this fresh connection doesn't log on again. This means the error keeps happening.

I added the resetClient and resetAllClients methods to the WebservicesAuthentication class to reset the sessionID before resetting the connection resulting in a fresh login.

### DIFF
--- a/src/Secure/WebservicesAuthentication.php
+++ b/src/Secure/WebservicesAuthentication.php
@@ -2,6 +2,7 @@
 
 namespace PhpTwinfield\Secure;
 
+use PhpTwinfield\Enums\Services;
 use PhpTwinfield\Exception;
 use PhpTwinfield\Services\BaseService;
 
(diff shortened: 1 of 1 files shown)


willemstuursma:
Looks good, finished or still a WIP?

fschalkwijk:
Short answer: It is still a work in progress.

It looked like it worked on my development and test servers, but when I deployed it to production the error kept occurring. I tried logging all requests to Twinfield confirming it tried to login again after an error. Which works. So either Twinfield is immediately revoking my session id or it is cached somewhere else. I haven't had time to check which is the case and if I can resolve it.

fschalkwijk:
Resolved WIP status.

My problem was that I had multiple queue jobs running concurrently. Apparently Twinfield doesn't like this. When I changed my queue worker to run just one job at a time, everything worked.

PR #156, opened 2019-09-20
Implemented connector for general ledger accounts

remkobrenters:


src/ApiConnectors/GeneralLedgerAccountsConnector.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public function listAll(
+        string $pattern = '*',
+        int $field = 0,
+        int $firstRow = 1,
+        int $maxRows = 100,
+        array $options = []
+    ): array {
+        $response = $this->getFinderService()->searchFinder(
+            FinderService::TYPE_DIMENSIONS_PROJECTS,
```
SuperRembo: If this connector is intended for general ledger accounts, then `TYPE_DIMENSIONS_FINANCIALS` would be more appropriate.

The results should also be limited to level 1 dimensions only.

remkobrenters: True about the dimension. Was lazy as they both reference to DIM. Will change that. Can you elaborate the dimension limit wish as I am not sure what you intend.

After I will submit a clean, one commit, PR for this change.

PR #157, opened 2019-09-30
Documentation

SuperRembo:
* Change all links to the Twinfield API documentation to use the current default cluster as the old (C3) cluster will be phased out.
* Some small spelling and grammar fixes in comments


willemstuursma:
Cheers

SuperRembo:
The project description still contains the c3 link. Would be good to change that too.

Issue #160, opened 2019-12-15
PeriodField on SalesTransaction should be able to be empty


niektenhoopen:
In PeriodField:
```
Period in 'YYYY/PP' format (e.g. '2013/05'). If this tag is not included or if it is left empty, the periodis determined by the system based on the provided transaction date.
```

But it is not allowed to be empty because this is how the period is set:
```
        if (!preg_match("!\\d{4}/\\d{1,2}!", $period)) {
            throw new \InvalidArgumentException("Period must be in YYYY/PP format (got: {$period}.");
        }
```

When I don't do setPeriod() on a SalesTransaction I get the following error:
```
TypeError : Argument 1 passed to PhpTwinfield\BaseTransaction::setPeriod() must be of the type string, null given, called in .../vendor/php-twinfield/twinfield/src/Mappers/TransactionMapper.php on line 88
```

PR #161, opened 2019-12-15
Transaction Period should be able to be empty

niektenhoopen:
Possible fix for #160 

It works for me but I only use sales transactions in my project. I believe you have a broader experience with different types of documents.

### DIFF
--- a/src/Transactions/TransactionLineFields/PeriodField.php
+++ b/src/Transactions/TransactionLineFields/PeriodField.php
@@ -26,7 +26,7 @@ trait PeriodField
      */
     public function setPeriod(?string $period): self
     {
-        if (!empty($period) && !preg_match("!\\d{4}/\\d{1,2}!", $period)) {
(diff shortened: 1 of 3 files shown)


src/Transactions/TransactionLineFields/PeriodField.php
```diff
@@ -24,9 +24,9 @@ public function getPeriod(): ?string
      * @param string $period
      * @return $this
      */
-    public function setPeriod(string $period): self
+    public function setPeriod(?string $period): self
     {
-        if (!preg_match("!\\d{4}/\\d{1,2}!", $period)) {
+        if (!empty($period) && !preg_match("!\\d{4}/\\d{1,2}!", $period)) {
```
willemstuursma: I prefer actual strict `null` checks, as `empty` can match all kinds of things in PHP (such as `"0"`).

willemstuursma:
Thanks, great stuff.

PR #164, opened 2020-01-03
Add option to select an office in Twinfield

### LINKED ISSUE #163: How do you set the office for the Browse data?
How do I define which office I want to browse data for?

lenvanessen:
This allows the user to select the office currently used in Twinfield. This is needed for the BrowseData endpoint. Also fixes #163 

https://accounting2.twinfield.com/webservices/documentation/#/FAQ


lenvanessen:
@willemstuursma can you review this and how do you feel about the approach?

lenvanessen:
Btw, If the approach is good I'll update the tests as well

willemstuursma:
Hi @lenvanessen. Thanks for contributing this PR.

I don't work for Twinfield so it took me some time to find time to look at this PR.

I feel it can be improved by merging the `SelectOfficeService` you created with the anonymous class in `\PhpTwinfield\Secure\WebservicesAuthentication::login`. I would then rename it the `SessionService`. 

Otherwise it looks fine. 

If you could make this change and ideally add some tests, I will make a timely release.

lenvanessen:
Sure thing, gonna work on it now. Wasn't sure about merging the two, because the session.asmx on login.twinfield.com is different from the one from the cluster accounting.twinfield.com, it's missing the SelectCompany method. But it does have the Logon method so I think we'll be able to merge the two.

Gonna update the PR soon!

willemstuursma:
Cool. Planning on adding any test cases? Can you confirm it all still works?

lenvanessen:
Just updated the test case, you might want to double-check them. I'm not that familiar with testing Soap, as when Soap was popular I was still in diapers;)

All test cases show no issues, and also manual testing shows it's properly switching the office as well as authenticating and fetching data.

PR #166, opened 2020-02-05
Allow reuse of an access token

axlon:
Right now the OpenIdConnectAuthentication always retrieves a fresh access token from Twinfield.

This tiny PR allows implementors of this library to manage access tokens, this means an application could cache and reuse a token (resulting in less requests made to Twinfield and thus faster response times).

I added the access token as an optional last parameter so there are no breaking changes.

### DIFF
--- a/src/Secure/OpenIdConnectAuthentication.php
+++ b/src/Secure/OpenIdConnectAuthentication.php
@@ -48,8 +48,9 @@ class OpenIdConnectAuthentication extends AuthenticatedConnection
      * Please note that for most calls an office is mandatory. If you do not supply it
      * you have to pass it with every request, or call setOffice.
      */
-    public function __construct(OAuthProvider $provider, string $refreshToken, ?Office $office)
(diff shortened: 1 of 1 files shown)


src/Secure/OpenIdConnectAuthentication.php
```diff
@@ -48,8 +48,9 @@ class OpenIdConnectAuthentication extends AuthenticatedConnection
      * Please note that for most calls an office is mandatory. If you do not supply it
      * you have to pass it with every request, or call setOffice.
      */
-    public function __construct(OAuthProvider $provider, string $refreshToken, ?Office $office)
+    public function __construct(OAuthProvider $provider, string $refreshToken, ?Office $office, string $accessToken = null)
```
niektenhoopen: Shouldn't this be `?string $accessToken` ?
```suggestion
    public function __construct(OAuthProvider $provider, string $refreshToken, ?Office $office, ?string $accessToken = null)
```

PR #171, opened 2020-02-19
Add like browse column operator

axlon:
Twinfield seems to return this operator when querying the browse definition for `130_2` (Customers v2)

Code to reproduce:

```php
$connector = new BrowseDataApiConnector(...);
$connector->->getBrowseDefinition('130_2');
```


willemstuursma:
Thanks!

PR #173, opened 2020-02-27
Allow dates from Browse data to be null

axlon:
Right now querying nullable dates from the browse API will cause an exception, in my case the matchdate on Supplier transactions (code 200)

The next line of code will cause an exception:
```php
(new BrowseColumn)->setField('fin.trs.line.matchdate')->setVisible(true),
```

Exception: `Invalid date, expected in yyyymmdd format, got "".`

With this PR no exception is thrown and a `BrowseDataCell` with value `null` is returned instead

I opted to not change the Util::parseDate and parseDateTime methods themselves, because they are used in other places as well.

### DIFF
--- a/src/Mappers/BrowseDataMapper.php
+++ b/src/Mappers/BrowseDataMapper.php
@@ -97,9 +97,9 @@ class BrowseDataMapper extends BaseMapper
             case 'Value':
                 return floatval($value);
             case 'Date':
-                return $value !== '' ?  Util::parseDate($value) : null;
+                return Util::parseDate($value);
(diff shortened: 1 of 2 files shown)


willemstuursma:
@axlon `parseDateTime` already support `null`, I think it is fine to add it to `parseDate` as well.

axlon:
@willemstuursma changed it in parseDate 👍

willemstuursma:
Thanks for checking the usages.

axlon:
@willemstuursma No problem! Could you tag a new release for this?

PR #175, opened 2020-03-02
Remove PerformanceType type from InvoiceLine

Wouter0100:
When this is set, mapping an invoice won't work in the InvoiceMapper. This because the InvoiceMapper maps a string to this method, which would result in the following error:
```
TypeError: Argument 1 passed to PhpTwinfield\InvoiceLine::setPerformanceType() must be an instance of PhpTwinfield\Enums\PerformanceType or null, string given, called in x/vendor/php-twinfield/twinfield/src/Mappers/InvoiceMapper.php on line 107 and defined in x/vendor/php-twinfield/twinfield/src/InvoiceLine.php:195
```

Besides that, this is consistent with all other Enums (for example, Status).

### DIFF
--- a/src/InvoiceLine.php
+++ b/src/InvoiceLine.php
@@ -187,12 +187,12 @@ class InvoiceLine
         return $this;
     }
 
-    public function getPerformanceType(): ?PerformanceType
+    public function getPerformanceType()
     {
         return $this->performanceType;
     }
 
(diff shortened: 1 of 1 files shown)


willemstuursma:
I feel it would be better to change the InvoiceMapper to cast the string to the correct type @Wouter0100. Any reasons not to do that?

PR #178, opened 2020-06-24
Allow custom transaction classes to use invoice number and payment reference

axlon:
When sending a custom transaction to Twinfield (one that extends one of the package transactions and doesn't directly use package traits), invoice number and payment reference will be skipped even if they are present. This PR fixes that by using the utility that already exists within the package.

The mapper side of things seems to be unaffected as it already uses this utility.


axlon:
@willemstuursma could you please take a look at this PR?

Neol3108 (approved):

willemstuursma:
Cheers.

PR #182, opened 2020-10-20
Update SalesTransaction.php

stepan-popovych:
We need to use Due Date for our SalesTransaction


stepan-popovych:
Need to fix null data DueDate for response

stepan-popovych:
Need to fix null data DueDate for response

PR #184, opened 2020-11-03
Change $messages default value to an empty array

rojtjo:
This prevents a `count(): Parameter must be an array or an object that implements Countable` warning on PHP 7.2 and above.


willemstuursma:
Cheers.

PR #185, opened 2020-11-19
Fixing issue where default was converted to 0 or 1 in the xml

rojtjo:
Address default attribute requires a string value of `true` or `false`. Currently booleans are converted to `0` or `1` when building the xml which makes the request fail.


rojtjo:
This might actually cause backward compatibility issues for people who insert `false` as a string since `"false" ? "true" : "false"` will always return `"true"`.

willemstuursma:
Cheers.

PR #168, opened 2020-02-15
getElementsByTagName searches elements on all levels

faisalmirza:
I was not able to retrieve a purchase transaction, it would fail on the fact that BNK line is fetched. This was my case: 

```
<?xml version="1.0" encoding="UTF-8"?>
<transaction location="temporary" result="1">
   <header>
      <office name="ABC BV" shortname="NL Sjabloon">OFFicecode</office>
      <code name="Inkoopfactuur" shortname="Inkoop">INK</code>
      <number>202000001</number>
      <period>2020/02</period>
      <currency name="Euro" shortname="Euro">EUR</currency>
      <regime>Generic</regime>
      <date>20200211</date>
      <origin>manual</origin>
      <originreference />
      <modificationdate>20200211093023</modificationdate>
      <user name="Site supervisor" shortname="Site supervisor">SUPER</user>
      <inputdate>20200211093023</inputdate>
      <duedate>20200312</duedate>
      <invoicenumber>1258962</invoicenumber>
      <freetext1 />
      <freetext3 />
   </header>
   <lines>
      <line type="total" id="1">
         <dim1 name="Crediteuren" shortname="Crediteuren" dimensiontype="BAS">1600</dim1>
         <dim2 name="Cooperatief U.A." shortname="" dimensiontype="CRD">2000</dim2>
         <debitcredit>credit</debitcredit>
         <basevalue>12100.00</basevalue>
         <rate>1</rate>
         <value>12100.00</value>
         <description>Factuur x</description>
         <vatbasetotal>2100.00</vatbasetotal>
         <vattotal>2100.00</vattotal>
         <matchstatus>matched</matchstatus>
         <matchlevel>2</matchlevel>
         <relation>2</relation>
         <matchdate>20200212</matchdate>
         <matches>
            <set status="temporary">
               <matchdate>20200212</matchdate>
               <lines>
                  <line>
                     <code>BNK</code>
                     <number>202000001</number>
                     <line>2</line>
                     <method>payment</method>
                     <matchvalue>12100.00</matchvalue>
                  </line>
               </lines>
               <matchvalue>-12100.00</matchvalue>
            </set>
         </matches>
      </line>
      <line type="detail" id="2">
         <dim1 name="Administratiekosten" shortname="" dimensiontype="PNL">4853</dim1>
         <debitcredit>debit</debitcredit>
         <basevalue>10000.00</basevalue>
         <rate>1</rate>
         <value>10000.00</value>
         <description>Factuur x adminkosten 2018</description>
[...]

PR #186, opened 2020-11-30
Fix TransactionMapper to ignore 'line' elements that are not actually transaction lines

axlon:
# Problem
Transactions sometimes have `<line>` elements that are not actually transaction lines (e.g. match lines), these lines will break the mapper as it tries to map them because information is missing.

# Solution
By using a XPath we only iterate over the `<line>` elements that are actually transaction lines

Closes #168


willemstuursma:
Cheers.

PR #194, opened 2021-05-10
Made sure the customer's setCountry function return the customer whic…

rickgoemans:
Made sure the setCountry function on the Customer class returns the customer which allows for method chaining.


willemstuursma:
Cheers, I see that indeed this had been omitted.

PR #195, opened 2021-06-01
Support PHP 8 and Money 4

bart-mollie:
This PR moves the supported version of the library to match the currently supported version of PHP:

- Support for 7.1 and 7.2 is dropped.
- Support for 8.0 is added.

Unfortunately `mediact/dependency-guard` does not have a PHP 8 compatible release yet, so I dropped it.
I've upgraded PHPStan and PHPUnit to versions that do support it.

I do understand dropping `dependency-guard` can be a deal breaker. I'd like you to consider allowing `moneyphp/money 4.0` in that case.

### DIFF
--- a/src/ApiConnectors/ArticleApiConnector.php
+++ b/src/ApiConnectors/ArticleApiConnector.php
@@ -55,11 +55,7 @@ class ArticleApiConnector extends BaseApiConnector
      */
     public function send(Article $article): Article
     {
-        $articleResponses = $this->sendAll([$article]);
-
-        Assert::count($articleResponses, 1);
-
-        return $articleResponses->getIterator()->current()->unwrap();
+        return $this->unwrapSingleResponse($this->sendAll([$article]));
     }
 
     /**
--- a/src/ApiConnectors/ArticleApiConnector.php
+++ b/src/ApiConnectors/ArticleApiConnector.php
@@ -59,9 +59,7 @@ class ArticleApiConnector extends BaseApiConnector
 
         Assert::count($articleResponses, 1);
 
-        foreach ($articleResponses as $articleResponse) {
-            return $articleResponse->unwrap();
-        }
+        return $articleResponses->getIterator()->current()->unwrap();
     }
 
     /**
--- a/src/Mappers/TransactionMapper.php
+++ b/src/Mappers/TransactionMapper.php
@@ -152,11 +152,11 @@ class TransactionMapper
                 ->setId($lineElement->getAttribute('id'))
                 ->setDim1(self::getField($transaction, $lineElement, 'dim1'))
                 ->setDim2(self::getField($transaction, $lineElement, 'dim2'))
-                ->setValue(Money::EUR(100 * self::getField($transaction, $lineElement, 'value')))
+                ->setValue(self::getFieldAsMoneyObject($transaction, $lineElement, 'value'))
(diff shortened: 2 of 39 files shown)


src/Mappers/TransactionMapper.php
```diff
@@ -263,6 +263,20 @@ private static function getField(BaseTransaction $transaction, \DOMElement $elem
...
+        \DOMElement $element,
+        string $fieldTagName
+    ): ?Money {
+        $fieldValue = self::getField($transaction, $element, $fieldTagName);
+
+        if ($fieldValue === null) {
+            return null;
+        }
+
+        return Money::EUR((string)(100 * $fieldValue));
```
willemstuursma, line 277: This EUR centricity is not alright. The method should take the currency as an argument.

src/Mappers/TransactionMapper.php
```diff
@@ -263,6 +263,20 @@ private static function getField(BaseTransaction $transaction, \DOMElement $elem
         return $fieldElement->textContent;
     }
 
+    private static function getFieldAsMoneyObject(
```
willemstuursma, line 266: I think that the `Object` postfix is unnecessary.

src/ApiConnectors/ArticleApiConnector.php
```diff
@@ -59,9 +59,7 @@ public function send(Article $article): Article
 
         Assert::count($articleResponses, 1);
 
-        foreach ($articleResponses as $articleResponse) {
-            return $articleResponse->unwrap();
-        }
+        return $articleResponses->getIterator()->current()->unwrap();
```
willemstuursma, line 62: This should be moved to a `final protected method unwrapSingleResponse(array $responses)` on the parent class.

src/Mappers/TransactionMapper.php
willemstuursma: The method is a great addition, though.

willemstuursma (changes requested):

src/Mappers/TransactionMapper.php
bart-mollie: I added currency as an argument. Do note that I changed the determination to default to EUR if nothing is set, and this default currency will now always be set on the transaction.

Note that tetermining whether the currency is set in the response is now done with an `isset` instead of `empty`. The tests didn't complain about it, but I'm not sure about possible edgecases missed.

src/Mappers/TransactionMapper.php
bart-mollie: ✅  Removed the postfix.

src/ApiConnectors/ArticleApiConnector.php
bart-mollie: ✅ Added the method to the base class and used it.

src/Mappers/TransactionMapper.php
```diff
@@ -263,18 +261,19 @@ private static function getField(BaseTransaction $transaction, \DOMElement $elem
...
+        Currency $currency
     ): ?Money {
         $fieldValue = self::getField($transaction, $element, $fieldTagName);
 
         if ($fieldValue === null) {
             return null;
         }
 
-        return Money::EUR((string)(100 * $fieldValue));
+        return new Money((string)(100 * $fieldValue), $currency);
```
willemstuursma, line 278: hmm this `100 *` is going to cause some issues, guess we should use `DecimalParser` but good enough for now for me.

willemstuursma (approved):

PR #197, opened 2021-07-15
feat: added list all ledger accounts connector

mfrankruijter:
I've added the Ledger Account connector, based on the documentation: https://accounting.twinfield.com/webservices/documentation/#/ApiReference/Masters/BalanceSheets
The code is similar/identical to that of the VatCodeApiConnector.

We require this functionality for a project.


willemstuursma:
LGTM

PR #198, opened 2021-08-18
Adding ability to fetch VAT code details

rojtjo:
For a project I'm working on I need the ability to lookup the VAT percentages. I've tried matching the coding style as much as possible, but it might be off in some places.

Also, I'm not quite sure about the naming I used for the objects as `VatCode` (less detailed) was already used for the `listAll` method on the connector.

Happy to change anything, let me know!

### DIFF
--- a/src/Account.php
+++ b/src/Account.php
@@ -1,0 +1,83 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpTwinfield;
+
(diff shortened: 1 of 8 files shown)


willemstuursma:
Thanks, I'll have a look tomorrow.

willemstuursma:
Thanks.

PR #199, opened 2021-08-25
Handle TypeError when dim1 is omitted on VAT lines

rojtjo:
I ran into a TypeError while omitting dim1 field on a transaction line with type VAT. According to the [Twinfield documentation](https://accounting.twinfield.com/webservices/documentation/#/ApiReference/Transactions/SalesTransactions) the dim1 field can be empty for these transaction lines.

> If line type = `vat` the VAT balance account. When an empty dim1 is entered, by default the general ledger account will be taken as entered at the VAT code in Twinfield.

This seems to be true for all transaction types which allow VAT lines.

### DIFF
--- a/src/DomDocuments/TransactionsDocument.php
+++ b/src/DomDocuments/TransactionsDocument.php
@@ -124,8 +124,11 @@ class TransactionsDocument extends BaseDocument
             $lineElement->setAttribute('id', $transactionLine->getId());
             $linesElement->appendChild($lineElement);
 
-            $dim1Element = $this->createNodeWithTextContent('dim1', $transactionLine->getDim1());
-            $lineElement->appendChild($dim1Element);
+            $dim1 = $transactionLine->getDim1();
+            if (!empty($dim1)) {
+                $dim1Element = $this->createNodeWithTextContent('dim1', $dim1);
+                $lineElement->appendChild($dim1Element);
+            }
 
             $dim2 = $transactionLine->getDim2();
             if (!empty($dim2)) {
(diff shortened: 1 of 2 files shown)


willemstuursma:
Please add an accompanying test case that would have caught this error.

edwin-easytrans:
Hi @rojtjo, this is a breaking change. I don't have dim1 on my total line to choose the debtor ledger of the customer, instead of providing it in the request. After this update i get the following error: "Dimensie 1 moet altijd worden ingevuld.//De dimensie 1 mag niet leeg zijn als dimensie 2 gevuld is., Dimensie op niveau 2 mag niet worden ingevuld omdat de dimensie op niveau 1 leeg gelaten is."

So this change is not only applicable to the VAT line, but also to the Totals line.
I hope you are able to correct this. Thank you.

rojtjo:
It's been a while since I added this change so I'll have to read up on this again.
If you happen to have time to create a pull request for this feel free, otherwise I'll try to have a look at this somewhere next week.

edwin-easytrans:
If you could have a look, that would be great. Thanks.

rojtjo:
Hi @edwin-easytrans,

I'm sorry for the delay, I've been really busy these last few weeks.

I just pushed a potential fix to the `fix/only-skip-setting-dim1-on-vat-line` branch. You should be able to install this using `composer require php-twinfield/twinfield:dev-fix/only-skip-setting-dim1-on-vat-line`.

After you confirm this fixes your problem I'll merge these changes and make a new release.

edwin-easytrans:
Hi @rojtjo, Thank you very much for the fix. I have tested the changes and it works great again.
Really appreciate your help!

rojtjo:
Great to hear!
I have created a new release (3.3.0) which includes these changes.

PR #210, opened 2022-05-23
Allow meansofpayment to be set on suppliers

rojtjo:
This pull request also includes an integration test for the suppliers which was missing.


willemstuursma:
Cheers.

Issue #214, opened 2022-08-17
Twinfield to cloudflare infrastructure ("Could not connect to host")


spoxies:
On 15 aug 2022 there was a [infrastructure change](https://wktaaeu.force.com/nlcommunity/s/article/Brief-technisch-onderhoud-Twinfield-Boekhouden-op-15-en-17-augustus?language=nl_NL) and on our env this resulted in a "Could not connect to host". 
However there was no real connection issue as running direct (non SoapClient) requests from PHP were all fine.

A SSL/Cipher problem was the first direction I took in debugging, but nothing conclusive.
But it would explain how the occurrence coincided with the Twinfield Cloudflare migration.

The exact cause of this issue remains a bit of a black box, [WDSL caching](https://stackoverflow.com/questions/4318870/soapfault-exception-could-not-connect-to-host) could be a factor.
But in my case, clearing cache and rebooting the server, disabling WDSL cache gave no relief.

Opening a [PR](https://github.com/php-twinfield/twinfield/pull/215) for `PhpTwinfield\Services\BaseService` to fix this issue as proposed [here](https://stackoverflow.com/a/50235119),
as it either bypasses or (hot)fixes the issue for now (if others experience the same issue).

Since the proposed fix also mentions misconfiguration of the SoapClient,
the real fix might need to be implemented elsewhere so this PR is a Hotfix.

Env: 

- nginx
- PHP  8.0.22

```json
{
    "require": {
        "php-twinfield/twinfield": "^3.0",
        "league/oauth2-client": "^2.4"
    }
}

```

Error:
```

  | object(PhpTwinfield\Exception)#25 (7) {
-- | --
  | ["message":protected]=>
  | string(25) "Could not connect to host"
  | ["string":"Exception":private]=>
  | string(0) ""
  | ["code":protected]=>
  | int(0)
  | ["file":protected]=>
  | string(129) "[..]/vendor/php-twinfield/twinfield/src/ApiConnectors/BaseApiConnector.php"
  | ["line":protected]=>
  | int(111)
```

RickJeroen:
Twinfield need to fix their soap:address location in https://api.accounting.twinfield.com/webservices/processxml.asmx?wsdl

> <wsdl:service name="ProcessXml">
<wsdl:documentation xmlns:wsdl="http://schemas.xmlsoap.org/wsdl/">Twinfield Process XML web service methods.</wsdl:documentation>
[...]

PR #215, opened 2022-08-17
[FIX] Explicitly set SoapClient location

spoxies:
Remediates "Could not connect to host" error as extensively described in issue #214. 
The error seemed to coincide with the Twinfield move to the cloudflare infrastructure.


rojtjo:
Thank you for your pull request!
It seems like this is no longer needed as you described in #214 so I'll close this for now.

PR #224, opened 2023-08-25
Update TransactionTypeApiConnector.php

zetno:
Fixes like these are missing on a couple places. 

This seems to have been fixed in OfficeApiConnector but the fix is missing in e.g. TransactionTypeApiConnector.


rojtjo:
Thanks!

PR #227, opened 2023-10-06
Fix more Project salestransactionline issues

RobinGeuze:
Apparently all fields on salestransactionline that are listed in the documentation as 'total' only, can also be set on 'detail' if the line has a project.


rojtjo:
Thanks!

Tagged as 3.5.2

PR #229, opened 2024-03-25
Map vat code and allow remove invoice line

Wimmie:
This pull request adds the following functionalities

- The ability to remove a invoice line
- The vat code will now be filled when a invoice is retrieved from Twinfield


rojtjo:
Thanks!

PR #230, opened 2024-09-18
Loosen `psr/http-message` version to allow `^2.0`

mrtus:
By loosening the constraints consumers will be able to upgrade their `psr/http-message` constraints too.

This change is untested ❗


rojtjo:
Thanks!

mrtus:
Thank you for the additions @rojtjo !

PR #231, opened 2024-12-20
Fix: added empty string instead of null

ruben450:
Fix for messages:

Deprecated: DOMDocument::createTextNode(): Passing null to parameter #1 ($data) of type string is deprecated


rojtjo:
Thanks!

Released as v3.7.1

PR #232, opened 2025-03-12
Fix implicitly null deprecation

JurianArie:
> Implicitly nullable parameter types are now deprecated. - https://www.php.net/releases/8.4/en.php


rojtjo:
Released as `3.8.0`.

Thanks!

PR #234, opened 2025-04-04
Added Currency and Currency rate support

php-bhosted:
Only adding new rates and creating currencies. No support for getting existing currencies and rates from Twinfield.

Also updated the readme.md for showing this new functionality in the table at the bottom.


rojtjo:
Thanks for your PR!
Could you add some tests and fix the phpstan issues?

php-bhosted:
Fixed the error. No tests added yet.

PR #237, opened 2025-09-26
Allow postingrules element and children to the sendDocument call for a customer

eelkevdbos:
**Feature**
Adds the postingrules to the sendDocument call for a customer by
- adding the appropriate models (e.g. CustomerPostingRule + Line)
- adding getters and setters on the Customer model
- adding additions to the CustomerDocument class to render them

**Bonus**
- Adds some development instructions


rojtjo:
Thanks!

Released as 3.9.0 :rocket:

PR #240, opened 2025-12-02
Add support for Retry-After header in BaseApiConnector

eelkevdbos:
Will make the api connector respect retry delay signalled by Twinfield via response headers.


rojtjo:
Reading up on the rate limits I don't expect large retry-after values, but I don't feel comfortable enabling this by default as it might cause some timeout issues in some contexts (HTTP or queue for example).

Could you make this opt-in through [`ApiOptions`](https://github.com/php-twinfield/twinfield/blob/master/src/ApiConnectors/ApiOptions.php)?

eelkevdbos:
> Reading up on the rate limits I don't expect large retry-after values, but I don't feel comfortable enabling this by default as it might cause some timeout issues in some contexts (HTTP or queue for example).
> 
> Could you make this opt-in through [`ApiOptions`](https://github.com/php-twinfield/twinfield/blob/master/src/ApiConnectors/ApiOptions.php)?

That is a valid concern. My current project is in CLI context, so I was a little naive with the implementation. Just for reference, when brute forcing the rate limit, I've seen header values around ~45s.

With respect to the opt-in implementation, the most robust option is probably to add an explicit option like `bool $useRetryAfter = false` or `bool $useRetryAfterHeader = false`. Do you have any preference in naming?

rojtjo:
Sorry, somehow I didn't get a notification for this.

Either is fine with me.

eelkevdbos:
We should be good to go 🚀

rojtjo:
I've made some minor changes:
- Added `$useRetryAfterHeader` as a constructor argument of `ApiOptions`
- Prevented retrying rate limited requests when `useRetryAfterHeader` is `false`, as that would mean they would be retried instantly
- Changed `setUseRetryAfterHeader` to return a new object to keep the `ApiOptions` immutable and match the rest of the class

Let me know if this still works for you or if you think something should be changed.

eelkevdbos:
Still works for me, thanks!

rojtjo:
Thanks!

Released as `3.10.0`

